PR #437, opened 2026-08-19
docs(writing-for-agents): note the direct-consumer disclosure trap

caneff:
When a separate consumer reads a disclosed file directly (code-review opens
CODING_STANDARDS.md by name, a hook reads a prompt file), disclosure means
moving the always-on import, not the file — leave the file at its path, drop
only the pointer that force-loaded it. Content that consumer must see then
legitimately lives both inline in the always-on core and in the full file; the
duplication is correct because the copies serve different load paths.

Earned from the progressive-disclosure pass (agent-skills#432, gridfind#604,
second-brain-v2#114): re-deriving the split from the ladder alone risks renaming
the file out from under the reviewer, or stripping the invariants from the full
file to avoid "duplicating" them inline. One sentence, no worked example — the
skill teaches not-bloat, so it shouldn't bloat.


caneff:
Closing unshipped — the note doesn't earn its place. It warns against a contrived, low-odds mistake (renaming/gutting CODING_STANDARDS.md), the split worked without it, and the real content is one fact about code-review, not a general writing principle already covered by the skill.

PR #582, opened 2026-09-07
flow: drop the ponytail plugin, keep only ponytail-audit

caneff:
Uninstalls the ponytail plugin (SessionStart hook injected the persona every session), strips its settings.json entries, removes it from the CLAUDE.md persona list, and detaches ponytail-audit / spec-standards-review from the plugin's ponytail-review skill (both already inline the tags).


caneff:
Superseded: plugin already removed on main; residual wording redone directly on main

Issue #900, opened 2026-09-19
worker-stop-alert: subagent launches are still read per-turn, so a message mid-review manufactures a false alert


caneff:
`worker-stop-alert.sh` builds its subagent set (`$launched` in the verdict jq program) from `$after` — the slice of the transcript since the last turn start — so a subagent launched in turn N and still out when a peer message arrives in turn N+1 is no longer counted as outstanding. The worker's next stop is then classified `silent` and the controller is told it stopped without reporting.

#886 fixed exactly this window for background shells and Monitor tasks by scanning `$tasks`/`$finished` over the whole transcript (`$all[].value`); `$launched` and `$returned` were left on the turn slice as pre-existing scope, out of that ticket. The correctness axis on #886 confirmed the subagent case by driving the real hook with a fixture: a reviewer still out across an inbound message alerts.

The fix is the same one-line move — source `$launched` and `$returned` from `$all` rather than `$after` — plus a test in the shape of `bg-across-turns` in `worker-stop-alert.test.sh`. Weigh it against #859's failure mode: an id that never resolves leaves the session `waiting` forever and suppresses every later true alert, so the teammate-report and hand-back shapes must all still clear a launch once widened.

Filed from the #886 round-1 correctness review (finding C1, second half).

## Blocked by

None — can start immediately.

caneff:
Ruling from #886's round-2 verification, which this ticket must build on: **widening the outstanding set to the whole transcript is not the fix.**

#886 tried exactly that for background shells and Monitor tasks, and the verification pass caught what it costs. A launched id that never reaches a terminal state holds the verdict at `waiting` for the rest of the session — the alert never fires again, however silently the worker stops. That is #820's failure mode hidden behind a benign verdict, the same shape #859 hit with teammate ids.

Measured over 252 transcripts in `~/.claude/projects`: **965 of 1985 launches (45%) never emit a terminal notification at all**, and no age cutoff separates a live task from an abandoned one (terminated p95 3601 s, p99 87572 s; never-terminated median 3397 s). Full numbers and method: `docs/research/2026-09-19-background-task-terminal-states.md`.

So `$tasks`/`$finished` stayed on the turn slice, and `worker-stop-alert.test.sh` now has a test that pins the bound (`a never-terminated job does not suppress a later genuine silent stop`). Widening `$launched` the same way would multiply the same suppression across subagents, where the hand-back shapes are already the harder half to resolve.

[...]

Issue #885, opened 2026-09-19
Spec: implement-spec and burndown rebuilt on the herdr lane (map #776)


caneff:
Spec for map [Multi-worker lane on herdr: implement-spec and burndown rebuilt](https://github.com/caneff/agent-skills/issues/776). Decisions live on the map's closed tickets and on the trial [#781](https://github.com/caneff/agent-skills/issues/781); this is the build. Grilled 2026-09-19 against #781's 69-item hurt list.

## Problem Statement

`burndown` and `implement-spec` are parked. Their Orca call sites were removed in #700 and the files now say that nothing in them runs until the multi-worker lane is rebuilt on herdr. So the only way to drive N workers is by hand: a controller session, a pasted brief, a frontier computed with `gh` and a regex, and a human holding the run's state in their head.

That was done twice under supervision, once as a spec run (sudokumaker-custom-constraints #366, 6 tickets) and once as a burn (10 landings plus 1 park). Both finished. Both produced a hurt list, and the hurt list says the same thing in a dozen voices: **the pieces of the lane are sound, and the coordination between them is unwritten.** Clumping was computed from ticket prose and two workers collided over a shared `#include`. The frontier was computed from a markdown section that 8 of 21 tickets did not have. The run's state lived in a per-repo append-only file with five earlier runs in it. A WSL restart renamed every session and every worker's brief carried a controller address that no longer existed. The liveness backstop fired five times and was wrong five times.

None of that is a defect in `/implement`, `implement-dispatch` or `merge-cleanup`. It is the policy layer that was never written down.

## Solution

`burndown` and `implement-spec` come back as two skills that are **policy over the `/implement` lane's own primitives** — `implement-dispatch`, `herdr agent`, `merge-cleanup`, the sessions socket — and restate none of their grammar.

`burndown § The loop` is the single home for the dispatch loop (#779). `implement-spec` is a nested run — a worker to the burn that dispatched it, a controller to its own slices — that points at that loop with overrides. A run is one controller session on any primary checkout's default branch, driving N workers through `implement-dispatch --repo`, refilling a slot as soon as one frees, until its candidate set is drained or two consecutive parks stop it.

The coordination the trial had to invent by hand becomes stated policy: how the frontier is computed when blocking edges live in prose, how clumps are computed from the include closure rather than from declared seams, what the controller trusts for liveness now that the stop alert is measured noise, where run state lives so a run survives a restart, what parks a clump, what escalates to Chris, and what a controller must verify before it rules.

[...]

Issue #883, opened 2026-09-19
flow/install.test.sh is red on main: install.sh exits non-zero and an empty claude/agents dir aborts the install


caneff:
## TL;DR

`flow/install.test.sh` is red on `main` and has been for some time. Two
failures, reproduced on a pristine `origin/main` checkout at `46319ce` and
identically on an unrelated feature branch, so neither is a regression from
any open PR:

```
FAIL install.sh exited non-zero
FAIL an empty claude/agents dir aborted the install
```

## Why this matters more than the failure itself

This repo has **no GitHub check gate**. `statusCheckRollup` on PR #882 was
empty, and `gh run list --branch main` shows only a "Clear in-review on
close" workflow, skipped. So `flow/install.test.sh` and the other scripts
under `tests/` are the entire gate, they run only when somebody runs them,
and two of them have been failing on `main` unnoticed.

Found while controlling #877: its worker flagged the red gate in its PR-up,
and the controller reproduced it on `main` before merging.

## What to build

Fix the install path so an empty `claude/agents` directory does not abort
the install, and so `install.sh` exits zero on a clean run. Read
`flow/install.sh` and the test's own setup before deciding which side is
wrong — the test may be asserting the correct behaviour against a broken
install, or the reverse.

## Acceptance criteria

- [ ] `bash flow/install.test.sh` passes on `main` with no FAIL lines
- [ ] The empty-`claude/agents` case is covered by a test that fails before
      the fix
- [ ] Whichever side was wrong (script or test) is stated in the PR body

## Seams under test

- `flow/install.test.sh` — the existing suite; it already names both cases.

caneff:
A second suite witnesses the same defect: `bash tests/all.test.sh` fails on `main` with

```
FAIL an empty claude/agents dir aborted the install
```

Same root cause as this ticket's second half, so it is not a separate bug — but it means two suites go red, and a fix here should be checked against both. Verified on a clean `main` at 5956a4c, 2026-09-19.

Issue #781, opened 2026-09-14
Trial run: one batch and one spec through the multi-worker shape on herdr


caneff:
Part of #776

## Question

Prove the run shape before the skills are written. By hand on the herdr lane: one controller session on this repo's primary checkout, `--slots 5`, drive a small `ready-for-agent` batch through the burndown shape (explore, clump, `implement-dispatch` per clump, wait per the decided contract, merge lines to Chris) and one spec's slices through the implement-spec shape (exploration task first, one PR per slice, end-to-end test and spec-level review at the end). Pick the targets when this is takeable; this repo by default. Chris watches in herdr and merges. Record what hurt, as Lane 3's #493 did; the spec's tickets are cut from this run.

caneff:
Trial plan (2026-09-14): spec half first on sudokumaker-custom-constraints #366 (Up to N), by hand from the brief in `docs/research/2026-09-14-multi-worker-trial-brief.md`; burn half on the same repo once #787 (`implement-dispatch --spec`) lands and the spec run has settled. Gridfind's spec was passed over: its second slice is blocked by the deliberately parked #293. Hurt list goes here, one comment per item.

caneff:
Spec run #366, hurt: the Controller brief says the session is in the primary checkout on the default branch, but the harness started it in a worktree (`.claude/worktrees/sudokumaker-custom-constraints/spectest`, branch `spectest`). `implement-dispatch --repo` makes it harmless for dispatch, but the implement skill's front-door rule (`git branch --show-current` picks worker vs dispatcher) would call this session a worker. /implement-spec needs to state where the controller runs, or check it.

caneff:
Spec run #366, hurt: #371 is a research ticket (one new `docs/research/*.md`, tagged docs-only by the exploration pass) but carries no `documentation` label, so `implement-dispatch` sent it heavy tier: TDD, a three-axis review and a PR for a prose doc. The exploration pass knows the size tag before dispatch; /implement-spec should either add the label for a docs-only ticket or pass the tier, instead of trusting a label the ticket author forgot.

caneff:
[...]

PR #902, opened 2026-09-19
worker-stop-alert: background shells, Monitor tasks and closed loops are not silent stops

### LINKED ISSUE #886: Stop hook: background shells, Monitor tasks and inbound-message turn starts belong in the waiting set
## TL;DR

A worker that is waiting for a long job to finish looks, to the alert system, exactly like a worker that has quit without saying anything. So the controller gets told "this worker stopped without reporting" when nothing is wrong. It has happened six times and been wrong six times.

## What to build

`worker-stop-alert.sh` decides whether a worker that just went idle is legitimately *waiting* or has *stopped silently*. Its "still out" set holds three shapes of subagent hand-back (`.origin.senderTaskId`, `<task-id>`, `<teammate-message>`). Three common, correct worker behaviours are in none of them:

1. **An outstanding background shell.** A worker commits its change, launches `just check-full` under a background Bash, and goes idle waiting — exactly what the box rules ask for.
2. **Outstanding Monitor tasks.** Same shape, different primitive.
3. **Already reported, then answered a message needing no reply.** The hook measures "reported since the last turn start", and an inbound peer message *is* a turn start. So every controller message that closes a loop rather than asking a question manufactures a false alert on the worker's next stop.

Add all three to the waiting set. A worker in any of those states goes idle without an alert.

## Why this blocks the burndown spec

Measured across the two trial runs on #781: **six alerts, six false, zero true.** Three distinct causes, all three above. The spec (#885, story 20) demotes the alert from backstop to hint precisely because of this ratio — but the hint is still the only signal that catches the real silent finish, and a signal that is always wrong trains the controller to skip the pane read that would catch the real one. The third cause is the worst: the more diligently a controller closes its loops, the more false alerts it manufactures.

## Acceptance criteria

- [ ] A worker idle with an outstanding background shell produces no alert
- [ ] A worker idle with outstanding Monitor tasks produces no alert
- [ ] A worker that reported, then received a message needing no reply, produces no alert on its next stop
- [ ] A worker that genuinely stops without reporting, with nothing outstanding, still produces the alert
- [ ] Each of the four cases above has a test that fails before the fix

## Blocked by

None — can start immediately.

## Seams under test

- `worker-stop-alert.sh`'s waiting/silent classifier, driven by a transcript fixture per case. The seam is the classifier's verdict for a given transcript, not the herdr prompt it sends afterwards.

caneff:
Closes #886

## What changed

The Stop hook's "still out" set gains the two primitives it was missing — a background shell (`backgroundTaskId`) and a Monitor task (`taskId`), each ending only at a status-bearing `<task-id>` notification or a `TaskStop`, since a Monitor's event notifications carry a task id and no status while it keeps running.

A delivered report now covers every later stop until the worker does something: an inbound peer message starts a turn, so a controller closing its own loop ("merged, sha X") was manufacturing an alert on the worker's next stop.

`OPERATIONS.md` restates the new waiting set; `docs/research/2026-09-19-background-task-terminal-states.md` records the measurement that decided how wide that set may be.

## Tests run

`bash flow/claude/hooks/worker-stop-alert.test.sh` → ALL PASS (29 cases, 6 new). `bash tests/all.sh` → only `flow/install.test.sh`, red on `origin/main` already and filed as #883.

Each new behaviour was checked as a witness by mutation: dropping `$unfinished`, the `<status>` filter, the `TaskStop` branch, the `$delivered` arm or the `$reported_at` arm each fails exactly the tests that claim it.

## Decisions made

**Round-1 findings, all ten with their disposition** (joins `dispositions-886.jsonl`):

- **S1** header contract wording and wrap — fixed, `974f4f6`.
- **S2** the `$delivered` ("reported this turn") verdict arm had no test cover once the stale-report fixture gained its `work` line — fixed, `974f4f6` (new test: a report inside this turn covers the stop even when work followed it).
- **S3** / **C2** the `work()` fixture emitted a `tool_result` whose id never matched its call, and used `$RANDOM` — fixed, `974f4f6`.
- **S4** `OPERATIONS.md` over-long line — fixed, `974f4f6`.
- **S5** `$launched/$returned/$unresolved` and `$tasks/$finished/$unfinished` are the same three-step shape twice — `disputed: the two sets resolve by different grammars (hand-back shapes vs a status-bearing notification or TaskStop) and change independently; the reviewer called it acceptable today.`
- **S6** over-engineering cut, `.input.shell_id` in the `TaskStop` branch — `disputed: the live TaskStop schema documents shell_id as an accepted deprecated alias for task_id. Omitting it costs a task stuck "out" forever, which silently suppresses every later alert; keeping it costs one alternation. Verified present in 39 real transcripts.`
- **S7** over-engineering cut, `$RANDOM` in `work()` — fixed, `974f4f6`.
- **P1** the "a report stands while nothing has happened since it" proxy is broader than the ticket's case 3 and weakens its AC4 — `disputed: no mechanical separator exists between "merged, thanks" and "fix the findings" — real peer messages carry only from/from-name/from-mode, no needs-reply marker. Measured evidence is 6 false alerts and 0 true across the two #781 runs, and #885 story 20 demotes the alert to a hint backed by the sweep. AC4's own case, a worker that never reported, still alerts.`
- **P2** / **C1** the background-task set was read from the current turn only, so a message arriving mid-job put the job back on the ground — fixed in `974f4f6` by scanning the whole transcript, then **reverted in 1508412** on the round-2 finding below. The subagent half (`$launched`, pre-existing) is filed as **#900**, which now carries the ruling.

**Round-2 verification, V1** — widening the set to the whole transcript removed its only bound: one id that never reaches a terminal state holds the verdict at `waiting` for the rest of the session, so the alert never fires again however silently the worker stops. Measured over 252 transcripts: 965 of 1985 launches (45%) never emit a terminal notification, and the lifetime distributions of those that do and those that don't overlap heavily, so no age cutoff separates them. Fixed in `1508412`: the bound stays at the turn, the across-turn test now pins the bound rather than the window, and #900 holds the liveness-evidence design that would cover both. The accepted cost is the inverse error — a message arriving mid-job restarts the turn, so that stop alerts while the job runs. A false alert costs a pane read; a suppressed one costs the signal.

**Codex adversarial pass** (comment on this PR) raised both known trade-offs and recommended the two directions already disproved — a task-status source the transcript does not carry, and a fresh `SendMessage` for every peer-started turn, which reinstates the 6-false / 0-true case. The controller rejected both. Its second recommendation also offered a third option no axis reached — an explicit, machine-identifiable loop-closure signal — which is now recorded on **#900** as a candidate design with Codex credited; #900 carries it alongside the liveness-evidence bound.

**Also not fixed here:** `flow/install.test.sh` is red on `origin/main` (#883), untouched by this diff.

## Last reviewed sha

`974f4f6`. The verification pass covered `1508412`'s subject as finding V1 and its own witness test, but `1508412` itself landed after that pass and was not re-reviewed.


caneff:
[codex] Thread ready (01a0ba7f-2c0a-79d3-a510-f98c71b7b988).
[codex] Turn started (01a0ba7f-2d34-7c10-984d-bceeb1c0fb21).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’m inspecting the branch diff and classifier tests adversari...
[codex] Running command: /bin/bash -lc 'git status --short && git diff --check origin/main...HEAD && git diff --find-r...
[codex] Command completed: /bin/bash -lc 'git status --short && git diff --check origin/main...HEAD && git diff --find-r... (exit 0)
[codex] Running command: /bin/bash -lc 'git log --oneline --decorate origin/main..HEAD && git show --format=fuller --s...
[codex] Command failed: /bin/bash -lc 'git log --oneline --decorate origin/main..HEAD && git show --format=fuller --s... (exit 1)
[codex] Running command: /bin/bash -lc "nl -ba flow/claude/hooks/worker-stop-alert.sh | sed -n '65,145p'; nl -ba docs/...
[codex] Command completed: /bin/bash -lc "nl -ba flow/claude/hooks/worker-stop-alert.sh | sed -n '65,145p'; nl -ba docs/... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: the classifier deliberately retains two path...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the classifier deliberately retains two paths that suppress or emit alerts contrary to the stated waiting/silent contract.

Findings:
- [high] An active job is dropped as soon as any peer message starts a new turn (flow/claude/hooks/worker-stop-alert.sh:72-75)
  `$tasks` and `$finished` are derived only from `$after`, which begins after the latest non-handback peer message. Thus a worker can launch `just check-full`, receive an unrelated controller message while it remains running, and then go idle; the task is absent from `$unfinished` and the hook alerts. The code comments explicitly acknowledge this false-alert path. That violates the requirement that an idle worker with an outstanding background shell or Monitor task not alert.
  Recommendation: Do not ship this as satisfying the outstanding-task acceptance criterion. Track task liveness across turn boundaries using a reliable task-status source or an explicit lifecycle marker, and add transcript tests for both shell and Monitor tasks that receive an inbound peer message while still active.
- [high] The prior-report exemption hides silent stops after any tool-free new turn (flow/claude/hooks/worker-stop-alert.sh:92-96)
  After a successful historical SendMessage, the classifier returns `reported` whenever there has been no later tool call or tool result. An inbound peer message is still a turn start, but a worker can receive a request, produce only assistant text (or stop without acting), and never successfully SendMessage a report. Under the hook's own contract that is unreported, yet this branch suppresses the alert because the turn used no tools. The implementation cannot distinguish a controller's loop-closing message from a request that happens not to require a tool.
  Recommendation: Limit report carry-forward to an explicit, machine-identifiable loop-closure signal, or require a fresh successful SendMessage for every ordinary peer-started turn. Add a test with a prior report followed by a peer request and a tool-free stop; it must alert.

Issue #915, opened 2026-09-20
Document the § cross-reference syntax the test gate enforces


caneff:
`tests/check-section-references.py` enforces a prose convention across every tracked Markdown file in this repo — which `§ <Heading>` forms resolve, that a bare `§` after a named file is a collision, that a heading only has to *start with* the referenced name, and (since #884) that `§ <Heading> step <n> ...` names the section and treats the locator and the prose after it as not part of the name.

None of this is written down anywhere an agent loads. The only statement of the `step <n>` form is a code comment above `STEP_LOCATOR`, and nothing says `§ Step 3` is not that form. An agent writing a new cross-reference has to read the checker to learn the grammar, or find out from a red suite.

Write the grammar into a doc agents actually read — `AGENTS.md` or a page under `docs/agents/` — and point the checker's module docstring at it.

Filed from the multi-axis review of #884 (finding S3).

## Blocked by

None — can start immediately.

Issue #914, opened 2026-09-20
reference_label has grown five prose trimmers with no stated order


caneff:
`reference_label` in `tests/check-section-references.py` now applies five unrelated English heuristics in sequence — split on `" and §"`, strip a trailing `and`, split on `'s`, split on sentence punctuation, and (new in #884) cut at a `step <n>` locator. Nothing states that the order matters, but it does: the step locator only works because it runs after the punctuation split, so `§ Before the PR: step 5` truncates at the colon first.

The sixth prose pattern will be another appended statement with the same unrecorded dependency. Make the trimmer set enumerable and its ordering contract explicit — a list of named trimmers applied in order reads better than a cascade, and the ordering becomes something a test can pin.

Evidence: `grep -n "value = " tests/check-section-references.py` shows the cascade; `§ Before the PR: step 5` at two sites in `implement/SKILL.md` is the reference whose label depends on the order.

Filed from the multi-axis review of #884 (finding S2).

## Blocked by

None — can start immediately.

Issue #913, opened 2026-09-20
The § step locator leaves three gaps the section-reference gate cannot see


caneff:
`reference_label` in `tests/check-section-references.py` now ends a reference name at a `step <n>` locator (`STEP_LOCATOR`), which fixed #884. Three gaps remain, all latent today — the suite is green and no tracked Markdown trips them:

1. **Step numbers are never checked.** `§ The merge step 6` reduces to the label `The merge`, so renumbering or deleting a step under a referenced heading keeps the gate green. Renumbering a list is exactly the drift this gate exists to catch. Cheap recovery: when `STEP_LOCATOR` matches, also require the target section to contain an `<n>.` list item.
2. **Only singular lowercase-or-capital `[Ss]tep N` matches.** `steps 3-5`, `STEP 3` and `step three` are missed, so a future `§ The merge steps 3-5 ...` re-creates #884's exact symptom — trailing prose swallowed into the section name, reported as a stale cross-reference. `grep -roE '§ [^§]{0,60}[Ss]teps? [0-9]'` over tracked Markdown finds 6 hits, all singular.
3. **A heading that is itself `Step N: ...` is still over-read.** The non-empty-name guard deliberately leaves `§ Step 3` alone, so `§ Step 3 and then some prose` keeps the prose in the label and fails against a real `## Step 3: ...` heading. `find-skills/SKILL.md` and `implement-paper/SKILL.md` carry such headings.

Each gap needs its own fixture under `tests/fixtures/section-references/`, wired into `tests/section-references.test.sh` the way `step-locator-valid` and `step-heading-name-invalid` are. Gaps 1 and 3 are also recorded as a known limit in the comment above `STEP_LOCATOR`.

Filed from the multi-axis review of #884 (findings C1, C2, P1).

## Blocked by

None — can start immediately.

Issue #887, opened 2026-09-19
implement: the PR-up report states the sha CLEAN was observed at, and accounts for every commit past the reviewed sha


caneff:
## TL;DR

When a worker says "my pull request is clean and ready", that can already be false by the time the controller reads it — the worker's own last push restarted the checks. And most workers report a final commit that their own review never looked at, leaving the controller to inspect it by hand.

## What to build

Two changes to the worker's PR-up report contract in `implement/SKILL.md`, plus one addition.

**1. CLEAN carries the sha it was observed at.** The report states the sha its `mergeStateStatus` reading came from. The controller's re-check at § The merge step 2 is then the only authority, and the staleness is explicit rather than a race the report silently loses.

**2. The tip is accounted for.** Either the tip equals the last reviewed sha, or the worker states the **diff class** of every commit past it — what kind of change it is, so the controller can decide whether it needs a review round without diffing it blind.

**3. A mutation check, when the deliverable is a test or a gate.** The report names one change that makes the new test or gate fail. Without it there is no evidence distinguishing a gate from a test that always passes.

## Why this blocks the burndown spec

From the burn on #781: **4 of 7 reports carried a tip past the reviewed sha**, and the controller diffed each one by hand to decide whether to accept it (#451 `ae83abf`, #453 `0ad42de`/`85fe6cb`, #454 `5867583`, #456 `ab1100e`). There is currently no stated rule for what a controller does with a trailing unreviewed commit, and the thing a controller reaches for under time pressure is to take the worker's word.

Separately, #456 reported CLEAN at `ab1100e`; seconds later the PR read UNSTABLE with `gate (check)` still pending. The CLEAN was true at `880aebb`, before the wording commit. Nothing landed wrongly — step 2 caught it — but it cost a controller wake and a round trip, and wakes are the scarce resource the fan-in design is built around.

The one report in the whole burn that cost the controller nothing (#501) did all three of these unprompted, including a mutation check nobody asked for.

## Acceptance criteria

- [ ] The report contract requires the sha at which CLEAN was observed
- [ ] The contract requires either tip == reviewed sha, or a stated diff class for each commit past it
- [ ] The contract requires a mutation check when the ticket's deliverable is a test or a gate
- [ ] `implement/SKILL.md`'s report template shows all three
- [ ] A guard test asserts the template carries each field

## Blocked by

None — can start immediately.

## Seams under test

[...]

PR #916, opened 2026-09-20
tests: end a § reference at its step locator

### LINKED ISSUE #884: tests/section-references.test.sh is red on main: the extractor reads trailing prose as part of a § reference
## TL;DR

`tests/section-references.test.sh` is red on `main`. Three cross-references
inside `implement/SKILL.md` name sections that do not exist in it:

```
implement/SKILL.md: § The merge step 3 Codex  not found in implement/SKILL.md
implement/SKILL.md: § The merge step 3 later writes its own files into this same `  not found in implement/SKILL.md
implement/SKILL.md: § The merge step 6  not found in implement/SKILL.md
```

Reproduced on a pristine `origin/main` checkout at `46319ce` and identically
on an unrelated feature branch, so this is not a regression from any open PR.

## What is actually wrong

Two different things, and the fix differs:

1. **The extractor over-reads.** Two of the three "missing" names are not
   section names at all — `§ The merge step 3 Codex` and `§ The merge step 3
   later writes its own files into this same \`` are the start of a prose
   sentence that happens to follow a `§` reference. The test is taking the
   rest of the line as part of the reference. That is a defect in the
   test's grammar, not in the document.
2. **`§ The merge step 6` may be a real stale reference.** `implement/SKILL.md`
   has a `### The merge` heading with numbered steps; whether step 6 still
   exists under that name needs a read.

Decide each on its own evidence rather than making the test pass.

## Why this matters

This repo has no GitHub check gate — `statusCheckRollup` on PR #882 was
empty. The scripts under `tests/` are the whole gate, they run only when
somebody runs them, and this one has been red on `main` unnoticed. A gate
nobody reads is worse than no gate, because the next real staleness arrives
indistinguishable from this noise.

Found while controlling #877, whose worker flagged it in its PR-up; the
controller reproduced it on `main` before merging.

## Acceptance criteria

- [ ] `bash tests/section-references.test.sh` passes on `main`
- [ ] The extractor stops taking trailing prose as part of a `§` reference,
      with a test case that fails before the fix
- [ ] Any genuinely stale reference is corrected in `implement/SKILL.md`
      rather than suppressed
- [ ] The PR body says, per finding, whether the document or the test was
      wrong

## Seams under test

- `tests/section-references.test.sh` — the existing suite and its reference
  extractor.

### LINKED ISSUE #912: tests/section-references.test.sh is red on main: sub-references like "§ The merge step 3" are rejected
`bash tests/all.sh` fails on `main` at `tests/section-references.test.sh` with three
errors, all in `implement/SKILL.md`: `§ The merge step 3 Codex not found`, `§ The merge
step 3 later writes its own files into this same ` not found`, and `§ The merge step 6
not found`. The section `## The merge` does exist — the checker
(`tests/check-section-references.py`) takes everything after `§ ` up to some delimiter as
the heading name, so a reference that points at a *step within* a section
("§ The merge step 3", "§ The merge step 6") never matches a heading.

Either the checker should accept a `step <n>` suffix after a real heading name, or those
references in `implement/SKILL.md` should be reworded — decide which after reading the
checker. Reproduce with `bash tests/section-references.test.sh` on a clean `main`; the
text came in with #882/#904 and `implement/SKILL.md` is byte-identical to `origin/main`.

Note this repo has no CI gate (see #883), so a third red suite on `main` went unseen the
same way the two in #883 did.

Filed from: the #883 build, which ran the full `tests/all.sh` gate and hit this
pre-existing failure outside its own diff.

caneff:
Closes #884
Closes #912

## What changed

`reference_label` in `tests/check-section-references.py` now ends a `§` reference name at a `step <n>` locator, so `§ The merge step 3 Codex pass ...` resolves to the section `The merge` instead of a heading nobody wrote.

Two fixtures under `tests/fixtures/section-references/` cover it: `step-locator-valid` (the bug) and `step-heading-name-invalid` (the hazard the fix introduces), both wired into `tests/section-references.test.sh`.

No Markdown document was changed. All three of the ticket's findings were false positives.

## Per finding: was the document or the test wrong?

The test, all three times. The ticket suspected the third of being a genuine stale reference; it is not.

| Ticket finding | Verdict |
| --- | --- |
| `implement/SKILL.md:158` — `§ The merge step 3 Codex` | **Test wrong.** The reference is `§ The merge step 3`; `Codex pass can't tell ...` is the sentence continuing. |
| `implement/SKILL.md:203` — ``§ The merge step 3 later writes its own files into this same ` `` | **Test wrong.** Same reference, more prose swallowed. |
| `implement/SKILL.md:221` — `§ The merge step 6` | **Test wrong.** `### The merge` runs steps 1–7; step 6 is the `gh issue view <n>` closure check, which is exactly what the citing sentence claims. |

One defect, three symptoms.

## Tests run

- `bash tests/section-references.test.sh` → `ok`, exit 0. Red at the base with exactly the three findings above — checked at both bases this branch has had, `d62b728` and the current `9a0d117` (line 217 became 221 when #887 landed; no fourth finding appeared).
- `python3 tests/check-section-references.py` (whole repo) → exit 0.
- `bash tests/all.sh` not used as evidence: it stops early on an unrelated failure in `tests/flow/install.test.sh` (#883, fixed on another branch).

## Decisions made

**No document was edited.** Every `§` reference in `implement/SKILL.md` is correct as written, so the fix is entirely in the extractor. Editing the three references to satisfy the test would have left the extractor broken and silently rewritten correct prose.

**A heading that is itself `Step 3: ...` is protected by two joint guards** in `^(.+?)\s+[Ss]tep\s+\d+\b` — the non-empty name and the mandatory whitespace before `step`. Measured against `step-heading-name-invalid`: `(.+?)\s+`, `(.*?)\s+` and `(.+?)\s*` all reject it; only `(.*?)\s*` accepts it, at which point the label is `""` and `matches_heading` accepts it against any heading. That fixture is the witness for the hazard, and it matters because `implement-paper/SKILL.md` and `find-skills/SKILL.md` carry ten `Step N:` headings between them.

**The grammar change was checked repo-wide.** Running both extractors over every `§` match in all 194 files the checker discovers, exactly three labels differ, and they are the three findings. No other reference's label moves, so nothing genuinely stale elsewhere stops being flagged. Re-run after the rebase onto #887's merge, since that commit added prose to `implement/SKILL.md`: still exactly three.

**Step numbers are now never validated** — `§ The merge step 6` checks only `The merge`. That is inherent to the fix the ticket asked for; filed as #913 rather than widened here.

**#912 is a duplicate of #884**, so this PR closes both. Filed under Chris's account at 04:31Z the same day, from the #883 build: same three errors, same file, same root cause, and it offers the same two options this build chose between — accept a `step <n>` suffix after a real heading name, or reword the references. This PR takes the first, which is the one it names first. Not extra scope; one fix, two tickets.

### Round-1 review findings

| Id | Disposition |
| --- | --- |
| S1 — the comment credited an inert guard | **Fixed** in `efdb399`, corrected again in `9409ab9`. The first fix credited the whitespace alone, which the verification pass caught as the same one-sided attribution in the other direction; measuring all four variants showed the two guards are joint, and the comment now says so. |
| S2 — `reference_label` is an accreting cascade with no stated ordering contract | **Filed** as #914. Restructuring the function is beyond this ticket's seam. |
| S3 — the `§ <Heading> step <n>` syntax is undocumented outside a code comment | **Filed** as #915. |
| P1 — a `Step N:` heading reference with trailing prose is still over-read | **Filed** as #913, and recorded as a known limit in the comment above `STEP_LOCATOR` in `efdb399`. |
| P2 — AC4 (the PR body states, per finding, whether document or test was wrong) unmet at review time | **Fixed** by this PR body's *Per finding* section above. |
| C1 — step-number staleness is now permanently unverifiable | **Filed** as #913. |
| C2 — the locator misses `steps 3-5`, `STEP 3`, `step three` | **Filed** as #913. Latent: `grep -roE '§ [^§]{0,60}[Ss]teps? [0-9]'` over tracked Markdown finds 6 hits, all singular. |

### Codex adversarial-review pass

One finding: `reference_label` reduces `§ The merge step 6` and `§ The merge step 8` alike to `The merge`, so a nonexistent step number now passes. Real, and word-for-word gap 1 of #913 — **filed** before the pass ran, duplicating this build's own C1 and P1. No code change, for three reasons.

Nothing is stale today: all three references resolve against the real `### The merge` and its 1–7 list, so no genuine staleness is being suppressed; the gap is future drift, which is what a tracked ticket is for. The catching power was accidental — before this change `§ The merge step 8` failed, but so did `§ The merge step 6` and every other step reference, and a gate that fails all of them distinguishes none of them. Going from always-red to green-with-a-tracked-gap is the right direction, and #913 names the cheap recovery precisely: when the locator matches, require an `<n>.` list item under the resolved section.

## Last reviewed sha

`efdb399`. One commit lands after it and was not re-reviewed: `9409ab9`, a comment-only correction to the round-1 S1 fix, raised by the verification pass. `git diff efdb399 9409ab9` touches comment lines only — `STEP_LOCATOR` and every other line of code are byte-identical across it.


caneff:
[codex] Thread ready (01a0bd1b-21b7-7853-bd36-28eeafb48971).
[codex] Turn started (01a0bd1b-228d-7370-8bca-bdd02289d9af).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’ll inspect the branch diff and test behavior adversarially,...
[codex] Running command: /bin/bash -lc 'git status --short; git rev-parse origin/main; git diff --check origin/main......
[codex] Command completed: /bin/bash -lc 'git status --short; git rev-parse origin/main; git diff --check origin/main...... (exit 0)
[codex] Running command: /bin/bash -lc "sed -n '1,260p' tests/check-section-references.py; sed -n '1,220p' tests/secti...
[codex] Command completed: /bin/bash -lc "sed -n '1,260p' tests/check-section-references.py; sed -n '1,220p' tests/secti... (exit 0)
[codex] Running command: /bin/bash -lc "sed -n '310,470p' implement/SKILL.md; rg -n '"'^#+ .*Step 6|''^6'"\\. |step 6'...
[codex] Command completed: /bin/bash -lc "sed -n '310,470p' implement/SKILL.md; rg -n '"'^#+ .*Step 6|''^6'"\\. |step 6'... (exit 0)
[codex] Running command: /bin/bash -lc "python3 - <<'PY' import importlib.util from pathlib import Path spec = importl...
[codex] Command completed: /bin/bash -lc "python3 - <<'PY' import importlib.util from pathlib import Path spec = importl... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: the new parser suppresses Step-N locator val...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new parser suppresses Step-N locator validity, so the gate can accept a stale numbered cross-reference.

Findings:
- [medium] Step number is discarded instead of validated (tests/check-section-references.py:68-69)
  `reference_label` reduces both `§ The merge step 6` and `§ The merge step 8` to `The merge`; the latter would pass whenever that heading exists. This makes the exact stale-reference case under investigation indistinguishable from a valid locator and violates the requirement to correct genuine staleness rather than suppress it. The new valid fixture covers only trailing prose, not an invalid locator number.
  Recommendation: Parse the optional Step-N locator as structured reference data and verify that the referenced numbered item exists under the resolved section. Add a failing fixture for a valid section with a nonexistent step number (for example, `§ Build step 8`) and a passing counterpart for an existing step.

Next steps:
- Preserve the step number through extraction and add locator-aware validation before shipping.

Issue #834, opened 2026-09-15
merge-cleanup: rebuild lane binaries when installed build is older than the last flow/lane commit, not only on HEAD movement


caneff:
Part of #776. Blocked by #823 (same file, merge_cleanup.rs).

`fast_forward_and_rebuild` rebuilds only when `git pull` moves HEAD during the cleanup itself. The controller's own trial-row auto-ship commit (implement skill § The merge step 3) pulls `main` first, so after a PR merges elsewhere, a later `merge-cleanup` run sees old_head == new_head, skips the rebuild, and runs a stale installed binary. Observed: after #829 merged, `merge-cleanup implement-821` skipped the rebuild and ran the pre-#821 binary (claim not cleared), fixed by hand with `bash flow/lane-install.sh`.

Rebuild when the installed binaries are older than the last `flow/lane` commit (e.g. compare against a recorded build sha), not on HEAD movement during this run.

Where: `flow/lane/src/bin/merge_cleanup.rs`, `fast_forward_and_rebuild`.

PR #917, opened 2026-09-20
flow: don't fail the install recording the build sha (#883)

### LINKED ISSUE #883: flow/install.test.sh is red on main: install.sh exits non-zero and an empty claude/agents dir aborts the install

caneff:
Closes #883

## What changed

`flow/lane-install.sh` recorded the build sha with `git rev-parse HEAD > "$state_dir/build-sha"`. In a checkout where HEAD does not resolve that fails under `set -euo pipefail`, so `install.sh` exited 128 after every symlink and both binaries were already in place — and the redirect still captured the literal `HEAD` that `rev-parse` echoes on failure, leaving `merge-cleanup` a garbage baseline.

The record now lives in a `record_build_sha` function the caller tests, so nothing in it (git, `mkdir -p`, or the write) can fail a completed install; a miss drops any stale record and warns, and `merge_cleanup.rs` rebuilds when there is no record.

`flow/install.test.sh` gains three cases — an unresolvable HEAD clears the record, a resolvable one records that sha, an unwritable `~/.local/state` does not fail the install — and one `scratch_repo` helper replaces four verbatim fixture copies.

**Which side was wrong: the script.** The `[ -e "$a" ] || continue` guard for an empty `claude/agents` was already on `main` and works; that case failed only because `install.sh` exited non-zero. One cause under both failures the ticket named, and it was in `flow/lane-install.sh`.

## Tests run

- `bash flow/install.test.sh` → `ALL PASS` (15 PASS, 0 FAIL; was 2 FAIL)
- `bash tests/all.sh` → every suite passes except `tests/section-references.test.sh`, which is red on `main` for an unrelated reason (`implement/SKILL.md` "§ The merge step 3" sub-references) and is filed as #912. `tests/all.test.sh` fails only by cascade from it; the `FAIL an empty claude/agents dir aborted the install` line the ticket's comment named is gone.

## Decisions made

Round-1 findings from `/multi-axis-code-review`, every one with its disposition:

- **S1** duplicated fixture setup — **fixed** in `34b9a18`: one `scratch_repo` helper for all four fixtures.
- **S2** compound assertion hides which half failed — **fixed** in `34b9a18`.
- **P1** AC3 (state which side was wrong in the PR body) — **fixed** here: see "Which side was wrong" above.
- **P2** clearing a stale record is beyond the ask — **disputed**: `merge_cleanup.rs` already treats a missing record as "no recorded build sha" and rebuilds, so leaving a stale sha is the unsafe option and no consumer change is needed.
- **P3** the positive-path recorded-sha case is adjacent scope from #834 — **disputed**: it is the paired half of the condition this change introduces; without it an inverted guard would silently stop recording. The verification pass proved it by strip-test.
- **P4** the fix lands at the shared root cause, not the `install.sh` agents loop the spec's prose named — **disputed**: the guard was already on `main`; the ticket invited deciding which side was wrong, and AC1 ("exits zero on a clean run") is the higher criterion.
- **C1** the record block could still abort the install (`mkdir`/write under `set -e`) — **fixed** in `34b9a18`, with a new case proven red against `0f62071`.
- **C2** recorded-case assertion swallows install output — **fixed** in `34b9a18`.
- **C3** a cargo failure would make the unresolvable-HEAD assertion blame the wrong thing — **disputed**: the adjacent `FAIL install.sh exited non-zero` fires in the same run and prints install.sh's whole output, so a toolchain failure is never hidden.

Outside the ticket's files, `tests/section-references.test.sh` was found red on `main` and filed as follow-up #912 rather than fixed here.

## Last reviewed sha

`34b9a18` — the verification pass covered round-1 findings and both fix commits. No commits after it.


caneff:
[codex] Thread ready (01a0bd1f-4f95-7333-a77b-aff1cdf56d95).
[codex] Turn started (01a0bd1f-505d-7972-a1dd-2c22c903c8fb).
[codex] Running command: /bin/bash -lc "git diff origin/main...HEAD -- flow/lane-install.sh flow/install.test.sh && pr...
[codex] Command completed: /bin/bash -lc "git diff origin/main...HEAD -- flow/lane-install.sh flow/install.test.sh && pr... (exit 0)
[codex] Running command: /bin/bash -lc "git show --stat --oneline 0f62071 && git show --format=fuller 0f62071 -- flow/...
[codex] Command completed: /bin/bash -lc "git show --stat --oneline 0f62071 && git show --format=fuller 0f62071 -- flow/... (exit 0)
[codex] Running command: /bin/bash -lc "nl -ba flow/install.test.sh | sed -n '1,220p'; printf '\\n--- tests all contra...
[codex] Command failed: /bin/bash -lc "nl -ba flow/install.test.sh | sed -n '1,220p'; printf '\\n--- tests all contra... (exit 2)
[codex] Assistant message captured: {"verdict":"approve","summary":"Ship: no material, change-induced correctness risk found in t...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: approve

Ship: no material, change-induced correctness risk found in the branch diff.

No material findings.

Issue #923, opened 2026-09-20
The brief still addresses the controller by Claude session name


caneff:
#892 put the herdr agent name in the run file, but the brief still carries a session name. `implement-dispatch` derives the controller address from the Claude session registry — its `--help` says "The controller is `--controller`, else the name in `~/.claude/sessions/<pid>.json` of the nearest ancestor process" — and interpolates that into the worker's `--controller "<name>"` (see `flow/lane/src/bin/implement_dispatch.rs`, grep `--controller`). A WSL restart renames every session, so every live worker is then holding an address that no longer exists: on the #781 spec run the controller went `spectest-1a` → `spectest-f3` and a worker `implement-368-42` → `implement-368-50`.

#892's own acceptance criterion says "Workers are addressed by herdr agent name **throughout**", and its What-to-build says "in the brief and in the run file". Only the run file half landed; `burndown/SKILL.md` now states the defect as motivation without fixing it. Raised as finding P1 by the spec axis reviewing the #892 build; parent spec #885, user story 25.

**Settled:** the design is resolve-at-send-time, not store-the-address. `SendMessage` cannot address a herdr agent name — every addressable peer is a Claude session name — but the herdr agent name **resolves** to one through two hops that both survive a restart: `herdr agent list` carries `agent_session.value` (the Claude session id), and `~/.claude/sessions/<id>.json` maps that id to the current `name`, which is what `SendMessage` takes. Measured on the #892 worker: herdr agent `skills-892` -> session `da4e24dc-8e2f-4d6c-a6a6-840a5907f441` -> `implement-892-47`.

So the run file records the herdr agent name (#892, done) and the address is resolved from it at the moment of sending. **Do not build the literal version** — writing the herdr agent name into `--controller "<name>"` produces an address `SendMessage` rejects. Baking a literal is what aged in the trial; a resolved address cannot age.

What this ticket covers: resolve the two hops at send time in the dispatch/controller path (`flow/lane/src/bin/implement_dispatch.rs`, grep `--controller`), the re-announce on resume, and the brief grammar wherever it states the controller address (`implement/SKILL.md`, `CONTEXT.md`).

Filed from the #892 review round (spec axis, finding P1).

## Blocked by

- #892

caneff:
Controller ruling, from measurement rather than preference — this resolves the blocking question this ticket was filed on.

[...]

Issue #893, opened 2026-09-19
burndown § The loop: dispatch, continuous refill, closure freshness, box check


caneff:
## Parent

#885

## TL;DR

The loop that drives several agents at once — pick the next jobs, start them, wait, start more as slots free up — has never been written down. It has been done twice by hand from a pasted set of instructions. This writes it as a skill.

## What to build

`burndown § The loop` — the **single home** for the dispatch loop (#779). `implement-spec` will later point at it with overrides and restate none of it.

**Where the controller runs.** Any primary checkout's **default branch**, not necessarily the target repo's; it addresses the target with `--repo`. The skill checks `git branch --show-current` against the default branch and refuses inside a worktree, where `/implement`'s front-door rule would call the session a worker.

**The candidate set is frozen** at the initial exploration, which covers the **whole queue** rather than the first wave. One exception: a ticket filed *during* the run because the run is stuck on what it fixes is dispatched into that run.

**Refill is continuous.** Recompute the frontier at each landing and dispatch into any free slot; no waves, because a wave holds slots empty waiting for its slowest clump.

**Closure freshness.** At each dispatch, re-resolve that clump's closure against current `main` — one hop — and check it against every in-flight workspace. Full re-exploration only after a landing whose diff touched a closure hub, because that is when other candidates' closures change.

**The frontier excludes** any clump whose closure intersects a live workspace. State the consequence out loud: a run drains **out of ticket order**, and on a repo with one hot shared file a single parked worker can hold a whole family. A controller reading only "open, unblocked, unclaimed" would dispatch into a collision.

**Box check before every dispatch** — `uptime` and `free -g` against the 28-process and ~24 GB rules.

## Evidence

From the burn on #781: after `#453` landed, the obvious next candidates were `#452`, `#457` and `#458` — all three in the `examples/_shared/line-kind.js` include family, the same family holding parked `#455`. Dispatching into it would have compounded the collision. The slot went to `#501` instead, outside `examples/` entirely. That is the exclusion rule working, and it is invisible to anyone reading the map's frontier definition.

## Acceptance criteria

- [ ] The skill refuses to run from a worktree, and says why
- [ ] The candidate set is frozen at the initial exploration, with the stuck-run exception stated
- [ ] The initial exploration covers the whole queue
- [ ] A freed slot is refilled without waiting for other clumps
- [ ] The frontier is recomputed at each landing
- [ ] A clump's closure is re-resolved at dispatch and checked against in-flight workspaces
[...]

Issue #891, opened 2026-09-19
Clump on the include closure, not on declared seams


caneff:
## Parent

#885

## TL;DR

Two agents were given tickets that looked unrelated and were told to work at the same time. They collided, because editing one shared snippet silently rewrites dozens of files that neither ticket mentioned. Work out what a change really touches before deciding two jobs are independent.

## What to build

The exploration pass groups candidate tickets into **clumps** — one worker per connected component of the file-collision graph. Today that graph is built from each ticket's declared seams, which is prose written by whoever filed the ticket.

Build it from the **include closure** instead: the set of files a change to the candidate's target files actually regenerates.

**How the closure is found.** The repo declares its include grammar and generator command in `AGENTS.md`. The exploration pass follows that declaration one hop from each candidate's target files.

**When the repo declares nothing.** Fall back to conservative clumping: any two candidates touching the same directory subtree are one clump. The run's opening report **says which mode it got**, so a controller can see whether it is being clumped precisely or conservatively.

**Never** resolve the closure empirically by touching files and running the generator per candidate. It is correct and far too expensive — a regeneration per candidate, per wave.

## Evidence

The costliest finding of the trial (#781). Ticket `#451` was "Line-kind gates cache the repeats answer in both directions"; its `## Seams under test` named each component's `update` and `validate` and a contract doc. Nothing about skyscraper. Read against `#455` (skyscraper components) and `#456` (outside-sudoku), the three looked disjoint and ran in parallel.

`#451`'s actual diff touched **41 files across six example families**, including the exact skyscraper component and four link artifacts that worker `#455` had already committed changes to. The cause was a shared `#include ../_shared/line-kind.js`: editing it regenerates every component that includes it, plus their generated links.

The ticket author could not reasonably have listed those files, and no amount of care reading the ticket would have found them. Cost: one rebase forced on an in-flight branch, which then hit a harness denial and parked that worker for hours.

## Acceptance criteria

- [ ] `AGENTS.md` carries a declared include grammar and generator command, and the resolver reads it
- [ ] The closure is resolved one hop from each candidate's target files
- [ ] Clumps are the connected components of the collision graph over closures
- [ ] A repo with no declaration falls back to directory-subtree clumping
- [ ] The fallback is announced in the run's opening report, in stated words
- [ ] A candidate whose files are included by nothing clumps alone
- [ ] No code path regenerates per candidate to find the closure

## Blocked by

None — can start immediately.

## Seams under test

[...]

PR #924, opened 2026-09-20
One run file per run, and workers addressed by herdr agent name

### LINKED ISSUE #892: One run file per run, and address workers by their herdr agent name
## Parent

#885

## TL;DR

If the machine restarts in the middle of a long run, the coordinating agent loses track of what it was doing and can no longer reach the agents it started. There is no record it can read to pick up where it left off.

## What to build

One **machine-readable file per run** at `~/.cache/burndown/<run-id>.json`.

**Contents.** The run id and slot budget; per clump, its ticket list, workspace path and herdr agent name; per landing, its squash sha.

**Why `~/.cache`.** It survives a WSL restart, which is what the standing rule against `/tmp` and the session scratchpad actually protects. It is not repo content, so it does not go into the target tree the workers branch from — and the controller may drive a repo it is not sitting on.

**Addressing.** Workers are addressed by their **herdr agent name**, in the brief and in the run file. A Claude session name does not survive a restart; the herdr agent name does.

**Resume.** On resume the controller reads the file, reconciles it against live state, and **re-announces itself to every live worker**, so a worker whose brief carries a dead controller address gets the current one.

The existing per-repo `~/.cache/burndown/<repo>.progress` file is retired.

## Evidence

From the spec run on #781: a WSL restart at 09:14 renamed every Claude session. The controller went from `spectest-1a` to `spectest-f3` and a worker from `implement-368-42` to `implement-368-50`. Every worker's brief hard-codes `--controller "<name>"`, so the next report would have gone to an address that no longer existed. The herdr agent name was the only stable handle.

Separately, the current progress file is per repo, not per run. It already held five earlier burns, each ending in `done`, before that run appended its first line — so a reader asking "is this run finished" or "which tickets are out" has to find where the run starts by eye.

## Acceptance criteria

- [ ] One JSON file per run at `~/.cache/burndown/<run-id>.json`
- [ ] It holds the run id, slot budget, per-clump ticket list / workspace / herdr agent name, and per-landing squash sha
- [ ] Resume reads it and reconstructs the run's state
- [ ] Resume re-announces the controller to every live worker
- [ ] Workers are addressed by herdr agent name throughout
- [ ] A run killed mid-flight and resumed recovers its slot budget, clump map and landing shas
- [ ] The per-repo `.progress` file is gone, and nothing reads it

## Blocked by

None — can start immediately.

## Seams under test

- The run file's read/write contract, round-tripped: write a run, kill it, resume from the file, assert slot budget, clump map and landing shas survive.
- The resume path's re-announce step, against a stub agent list holding both a live and a vanished worker.

caneff:
Closes #892

## What changed

`burndown/runfile.py` holds one run's state as one JSON file at `~/.cache/burndown/<run-id>.json` — run id, slot budget, the controller's herdr agent name, and per clump its ticket list, workspace, worker's herdr agent name and squash sha once it lands. `resume` reads it back and splits the clumps into the live workers to re-announce the controller to, the vanished ones to reconcile by hand, and the landings already banked, recovering the slot budget as it goes.

A worker is addressed by its herdr agent name throughout the run file, because a WSL restart renames every Claude session. `~/.cache` and not `/tmp`, because a restart wipes that and a restart is the event this file exists for; every write replaces the file in one step, because a torn run file reads as a run with no clumps.

The retired per-repo `~/.cache/burndown/<repo>.progress` log is gone — nothing in the tree reads or writes one, and `burndown/run-file-retirement.test.sh` keeps it that way while asserting `burndown/SKILL.md` points at the reader that replaced it. `references/run-file.md` holds the contract and the resume procedure.

## Tests run

- `python3 burndown/runfile_test.py` → `43 passed`
- `bash burndown/run-file-retirement.test.sh` → `PASS`
- `bash tests/all.sh` (the merge gate) → `52 suites passed`

Every constraint in the diff is mutation-checked: each was stripped, seen to fail, and restored. Two whose first mutation passed had their witnesses grown until they caught it (`makedirs` hoisted out of the `try`, and a `clumps` value that is not a list at all).

## Decisions made

Round-1 findings, each with its disposition (sidecars: `~/.cache/agent-reviews/skills/findings-{standards,spec,correctness}-892.jsonl`, dispositions in `dispositions-892.jsonl`):

- **S1** — fixed in `fe4f269`. The hollow witness is deleted; `os.replace` consumes the temp file on the success path, so its assertion could not fail. Its neighbour, which kills the write mid-flight, is the real witness.
- **S2** — fixed in `fe4f269`. One seam per caller, as `cost.py` does it: `BURNDOWN_CACHE_DIR` is read once in `main`, an in-process caller passes `root`.
- **S3** — fixed in `fe4f269`. The module docstring keeps the usage block, the shape and the pointer; the *Why* lives once, in `references/run-file.md`.
- **S4** — fixed in part in `fe4f269` (`checked` → `checked_run_id`, an identity comprehension became a `.split()`); `disputed: clump() doing validate / lookup / change / save in one function is the sibling convention — land() and closure.py have the same shape — and the overlap special case is one named variable.`
- **P1** — filed as **#923** (`ready-for-agent`, blocked by #892). `implement-dispatch` still derives the controller address from the Claude session registry and interpolates it into the brief's `--controller "<name>"`, so a restart still leaves live workers holding a dead address. That is a Rust change to live lane infrastructure past this ticket's state-file scope. Since filing, the controller measured the mechanism and #923 carries it as settled: `SendMessage` cannot address a herdr agent name, but the name **resolves** to one through `herdr agent list` → `~/.claude/sessions/<id>.json` → the current session name, and both hops survive a restart. So the run file stores the herdr agent name — the durable key, which is what landed here — and the address is resolved at send time, never stored.
- **P2** — `disputed: nothing here sends the re-announce because only the controller session can send a message; the run file's job is to name who to send to. The reviewer itself ruled the intent preserved, and the ticket's second seam asks for a stub agent list, which taking --live is what makes possible.`
- **P3** — `disputed: both invariants guard the run file's own answer — which tickets are out — rather than clumping policy.`
- **C1** — fixed in `fe4f269`. `os.makedirs` is inside the `try`, so a full or read-only `$HOME`, or a file where the cache dir belongs, is one stderr line rather than a stack trace.
- **C2** — fixed in `fe4f269`. The directory sync is `fsync_dir`, best effort by name and docstring: by the time it runs the write has landed, and reporting it as failed sent the controller back to a `start` that now refuses.
- **C3** — fixed in `fe4f269`. `load` validates each clump entry and the `clumps` value's own type, not just the four top-level keys.
- **C4** — fixed in `fe4f269`. A re-registration may grow a clump and never drop a ticket out of the run.
- **C5** — fixed in `fe4f269`. `parse_tickets` takes ASCII digits only.
- **C6** — `disputed: a run has one controller and one writer; references/run-file.md states that as a lane constraint rather than a guarantee this file enforces, and a second start on a live run id is refused.`

**Codex adversarial pass** (https://github.com/caneff/agent-skills/pull/924#issuecomment-5747961123), four findings:

- **F1** [high] — fixed. A vanished clump's slot counted as free, so a controller refilled into it before anyone had reconciled it — while its worker may still be holding those tickets. `reconcile` now reports `held` (live **and** vanished) and `free` as the budget minus that; only a landing frees a slot for certain. This is what `references/run-file.md` already said and the code did not.
- **F2** [high] — `disputed: "resume never re-announces" is true of this module and not a defect in it. Sending is SendMessage, a controller action; a Python module cannot perform it, and reconcile returning the announce bucket is the division — the run file supplies who to re-announce to, the controller does the announcing. The loop that performs it is #893; resolving each herdr agent name to a current address (it cannot be stored, only resolved) is #923. Nothing re-announces after a restart today because nothing runs a burn today: this skill is parked. Same disposition as #891's "not wired into a runnable exploration path", for the same reason.` The reference doc now states it where the resume procedure is described, so the next reader meets the decision.
- **F3** [high] — fixed. Every command that changes the file holds an advisory `flock` on `<run-id>.json.lock` across its read-modify-replace, with the same 30-second ceiling the dispatch lane waits with. Without it a landing's sha could be overwritten by a stale `landed: null` from a concurrently-loaded clump update, and a resumed controller would re-dispatch work that had already landed with a merged PR behind it. The witness runs three writers at once with their critical sections held open past each other's reads — the same device `flow/lane`'s `two_concurrent_dispatches_both_leave_their_trust_keys` uses — and a second witness holds the lock from outside and asserts the refusal, bounded so a waiter that never gives up fails the suite instead of hanging it.
- **F4** [medium] — fixed. `load` checked the keys and the `clumps` type and nothing else, so `slots: "2"` reached arithmetic in `resume` as a `TypeError`. It now checks every field: the run id against the filename it came from, the slot budget, the controller and each clump's workspace and agent as non-blank names, each ticket as a positive number, each landing as a git object name.

**Codex round 2** (https://github.com/caneff/agent-skills/pull/924#issuecomment-5748027796), two findings:

- **R2-1** [medium] — fixed. `BURNDOWN_RUNFILE_LOCK_TIMEOUT=30s` raised an uncaught `ValueError` from the `float()` in `locked`, so a value inherited from a parent shell turned a restart recovery into a traceback. Round 1's F4 was malformed *persisted state* escaping the clean-refusal path and this was malformed *environment config* escaping it one round later — the same contract, a new surface — so the whole boundary is swept rather than the named line: every environment read now goes through one guarded conversion. Unset or empty means the documented default (`VAR=` is how the shell clears an override); `30s`, `inf`, `nan` and `-5` are each refused by name, `inf` because it would wait forever and `nan` because it compares false against every deadline, which is the same wait without a name. That covers `BURNDOWN_RUNFILE_DELAY_MS` too, which had the same shape and which being test-only is an argument about likelihood rather than about the contract.
- **R2-2** [high] — `disputed: "resume never sends", re-raised. The dispute stands unchanged — runfile.py cannot call SendMessage and the controller can; reconcile returning the announce bucket is the division of labour. Round 2's sharper recommendation, a stub-messenger seam test, is a good one and is now #893's acceptance criterion (https://github.com/caneff/agent-skills/issues/893#issuecomment-5748036288): one message per live, unlanded worker, against a stub messenger. Captured where it can be built rather than disputed a third time.`

Two defects the round-1 fix commit introduced, both caught by the verification pass and fixed in `98cbbf1`, neither a review finding: `str.isdigit()` is true for `²` (a traceback) and for `٩` (which `int()` reads as 9, registering a ticket the brief never named), and resolving the cache-dir override in `main` had dropped its `expanduser`, so a tilde made a literal `./~/` directory.

Outside the ticket's named files: `CONTEXT.md` gains one **Run file** term, in the established term/_Avoid_ shape — the diff introduces a domain noun the prose in `burndown/SKILL.md` and the reference doc both use.

## Last reviewed sha

`fe4f269` — one round of `/multi-axis-code-review` (standards, spec, correctness) plus one verification pass. Three commits landed after it and were not re-reviewed by those axes: `98cbbf1` (the two verification-pass fixes), `4b99388` (Codex F1, F3, F4) and `75f2413` (Codex round 2, the environment boundary).


caneff:
[codex] Thread ready (01a0bd5a-e7e1-75f1-8bd7-9805fe7886fb).
[codex] Turn started (01a0bd5a-e892-70e0-b6dc-c8126c0d218b).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’ll inspect the branch diff and exercise the run-file contra...
[codex] Running command: /bin/bash -lc "sed -n '241,520p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md ...
[codex] Command completed: /bin/bash -lc "sed -n '241,520p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md ... (exit 0)
[codex] Running command: /bin/bash -lc "nl -ba burndown/runfile_test.py | sed -n '1,700p' printf '\\n--- complete diff...
[codex] Command completed: /bin/bash -lc "nl -ba burndown/runfile_test.py | sed -n '1,700p' printf '\\n--- complete diff... (exit 0)
[codex] Running command: /bin/bash -lc "nl -ba burndown/runfile_test.py | sed -n '420,640p' printf '\\n--- hook contro...
[codex] Command failed: /bin/bash -lc "nl -ba burndown/runfile_test.py | sed -n '420,640p' printf '\\n--- hook contro... (exit 1)
[codex] Running command: /bin/bash -lc "python3 - <<'PY' import sys sys.path.insert(0, 'burndown') import runfile for ...
[codex] Command completed: /bin/bash -lc "python3 - <<'PY' import sys sys.path.insert(0, 'burndown') import runfile for ... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: resume can over-dispatch unresolved work, do...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: resume can over-dispatch unresolved work, does not itself re-announce workers, and concurrent updates can erase landing state.

Findings:
- [high] Vanished clumps are incorrectly treated as free slots (burndown/runfile.py:242-254)
  `free` subtracts only live workers. A vanished clump is explicitly not yet reconciled and the run-file contract says it is not a free slot; nevertheless, resume with one live and one vanished clump in a three-slot run reports two free slots. The controller can dispatch replacements before resolving the vanished workspace, exceeding the run budget and potentially duplicating work.
  Recommendation: Count both live and vanished, unlanded clumps against the slot budget until the controller explicitly parks or re-dispatches the vanished clump.
- [high] Resume only prints re-announcement instructions; it never re-announces (burndown/runfile.py:257-265)
  The stated recovery requirement is to re-announce the current controller to every live worker. This function updates the JSON controller field and returns entries for rendering, but no changed code sends a message to the worker agent names. The only repository references to `runfile.py` are documentation. After a restart, workers therefore retain the dead controller address unless a human or an unimplemented future controller interprets stdout and performs extra work.
  Recommendation: Integrate resume with the controller's worker-prompt operation and make successful delivery to each live herdr agent part of the resume path, with failed deliveries surfaced for reconciliation.
- [high] Read-modify-replace updates can silently lose a recorded landing (burndown/runfile.py:194-214)
  `clump` loads the whole run, mutates it, and replaces the file without a lock or version check. If a landing command loads the same pre-update file and saves its SHA before this clump operation saves, the later clump replace writes its stale in-memory `landed: null` back to disk. A resumed controller then sees the landed work as in flight and can duplicate it. The reference admits the race but relies on an unenforced one-command-at-a-time convention, which is unsafe across CLI processes or re-entrant recovery actions.
  Recommendation: Serialize per-run mutations with an advisory lock, or add compare-and-swap/versioned writes that reload and retry when the run file changed.
- [medium] Malformed persisted state can escape the one-line recovery error path (burndown/runfile.py:124-143)
  `load` validates only required keys and whether `clumps` and `tickets` are nonempty lists; it does not validate `run_id`, `slots`, controller, ticket element types, workspace/agent strings, or `landed`. For example, a persisted `"slots": "2"` passes `load` and causes `reconcile` to raise an uncaught `TypeError` at the subtraction. This defeats the documented guarantee that an unrecognized run-file shape is refused cleanly, precisely when recovery needs a usable diagnostic.
  Recommendation: Fully validate the persisted schema and field invariants in `load`, including that the file run_id matches the requested filename; convert all invalid states to `RunFileError` before resume or mutation.

Next steps:
- Fix slot accounting and persist an explicit reconciliation outcome for vanished clumps.
- Make re-announcement an executed, observable resume operation.
- Add concurrency and malformed-file regression tests.

caneff:
[codex] Thread ready (01a0bd69-ffe8-7521-9086-733bf7a9057c).
[codex] Turn started (01a0bd6a-006f-7a20-afd8-999828572115).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’m inspecting the branch diff and recovery contract adversar...
[codex] Assistant message captured: {"verdict":"approve","summary":"I’m using the repository’s multi-axis review procedure becaus...
[codex] Running command: /bin/bash -lc "git diff --find-renames --unified=80 origin/main...HEAD -- burndown/runfile.py...
[codex] Command completed: /bin/bash -lc "git diff --find-renames --unified=80 origin/main...HEAD -- burndown/runfile.py... (exit 0)
[codex] Running command: /bin/bash -lc "nl -ba burndown/runfile.py | sed -n '80,464p'; nl -ba burndown/runfile_test.py...
[codex] Command completed: /bin/bash -lc "nl -ba burndown/runfile.py | sed -n '80,464p'; nl -ba burndown/runfile_test.py... (exit 0)
[codex] Running command: /bin/bash -lc "rg -n '"'^def test|''^def main|assert|LOCK_TIMEOUT|extra|duplicate|sorted|unre...
[codex] Command completed: /bin/bash -lc "rg -n '"'^def test|''^def main|assert|LOCK_TIMEOUT|extra|duplicate|sorted|unre... (exit 0)
[codex] Running command: /bin/bash -lc "git grep -n -E 'runfile[.]py|runfile[.]|runfile import|re-announce|SendMessage...
[codex] Command completed: /bin/bash -lc "git grep -n -E 'runfile[.]py|runfile[.]|runfile import|re-announce|SendMessage... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: resume records and prints a re-announcement ...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: resume records and prints a re-announcement plan but never re-announces the controller to workers, so the restart failure this change targets remains unresolved.

Findings:
- [high] Resume never sends the new controller address to live workers (burndown/runfile.py:345-356)
  `resume()` only persists the supplied controller name and returns a classification; the CLI renders `re-announce` lines. No changed controller/dispatch code consumes those lines or calls the worker messaging mechanism. After a restart, every live worker can therefore retain the dead controller address in its brief and continue reporting to it. The existing test proves only that text is printed, not that any worker receives an update.
  Recommendation: Wire resume into the controller path that resolves live herdr agents and sends each one the current controller address; add a seam test with a stub messenger that asserts one message per live, unlanded worker.
- [medium] An invalid lock-timeout override crashes recovery with a traceback (burndown/runfile.py:95-97)
  The timeout environment variable is converted with `float()` before the code enters a `RunFileError` boundary. A malformed inherited `BURNDOWN_RUNFILE_LOCK_TIMEOUT` (for example, `30s`) raises `ValueError`, bypassing `main()`'s error handling. This turns a restart recovery attempt into an unstructured traceback despite the documented one-line-refusal contract.
  Recommendation: Parse and validate the timeout inside a guarded conversion, reject non-finite or negative values as `RunFileError`, and add a CLI test for a malformed override.

Next steps:
- Implement and test actual controller re-announcement before shipping.
- Harden invalid lock configuration handling.

Issue #899, opened 2026-09-19
Unpark burndown and implement-spec, one supervised end-to-end run, finish line


caneff:
## Parent

#885

## TL;DR

Turn the two rebuilt skills back on, prove them with one real supervised run, and close the project.

## What to build

**Unpark.** `burndown/SKILL.md` and `implement-spec/SKILL.md` both currently say that nothing in them runs until the multi-worker lane is rebuilt on herdr. Remove those markers, and every stale pointer left from the parking commit (#752).

**Prove it.** One end-to-end supervised run on this repo: a small `ready-for-agent` batch through the burn shape, and one spec's slices through the nested shape. Chris watches in herdr. The run's own hurt list, if any, is recorded on this ticket.

**Finish line.** One command that checks both skills are live, no parked marker survives anywhere, and the run file, frontier reader and closure resolver are all reachable from the skills.

Map #776 closes when this lands.

## Acceptance criteria

- [ ] No parked marker remains in `burndown/SKILL.md` or `implement-spec/SKILL.md`
- [ ] No doc still points at the parked state
- [ ] One supervised burn run completes on this repo
- [ ] One supervised nested spec run completes
- [ ] Anything that hurt is recorded on this ticket
- [ ] The finish-line check is one command and passes
- [ ] Map #776 is closed

## Blocked by

- #889, #890, #891, #892, #893, #894, #895, #896, #897, #898 — the whole spec

## Seams under test

- The finish-line check itself: a grep over the skills tree for parked markers, plus a reachability check from each skill to the run file, frontier reader and closure resolver.
- The supervised run — the only seam that exercises the loop against real workers rather than stubs.

caneff:
## Hurt list — build side

The run's own hurt list is the controller's, from the two supervised runs. These four are what hurt building the unpark itself; recording them here per the acceptance criterion.

**1. Nothing in the repo said what parks a skill, so the brief guessed wrong.** The dispatch brief named `disable-model-invocation: true` as a parked marker and told me to cut it from both skills. It is in 53 of 79 `SKILL.md` files, `implement`'s among them, and it predates the parking commit `207fee4` — which touched only `description:`. Cutting it would have made two multi-worker run skills auto-invocable off model inference. Cost: one round-trip to the controller before the edit. Closed by `AGENTS.md` § Skill frontmatter, which now defines parked state and the property any detector of it must have.

[...]

Issue #779, opened 2026-09-14
Which file owns the shared dispatch loop, and what does each skill keep?


caneff:
Part of #776

## Question

Two skills share one dispatch loop (list frontier, check slots, explore and clump, dispatch, wait, hand off merge line, re-list). Which file owns the loop, and what does each skill keep? `implement-spec` keeps the spec-level steps (exploration task first, file-ownership brief, end-to-end test over the spec, spec-level review at the end); `burndown` keeps the mixed-queue rules (spec handoff, two-park stop, cost lines). Options: the loop lives in `burndown` and `implement-spec` points at it; the loop is a reference file both point at; `implement-spec` is `burndown` restricted to one spec's children plus its four steps. Decide, and decide where `spec-handoff.md` and `merge-tail.md` live.

caneff:
## Resolution

**burndown owns the loop; implement-spec is a nested run that points at it.** Grilled 2026-09-14.

1. **Nested, not inline.** When a burn meets a slice whose parent carries the `spec` label, it dispatches one worker whose brief is `/implement-spec <spec> --slots k --controller "<burn>"`. That worker is a Controller to the slice workers and a Worker to the burn: it reports to the burn as one opaque unit (PR set or park), exactly as the old spec handoff did. Chosen over inline for opacity (the burn never sees a slice's wake) and identity (implement-spec stays a skill someone starts).
2. **File ownership.** `burndown/SKILL.md` § The loop is the single home of the shared loop. `implement-spec/SKILL.md` says "run burndown § The loop with these overrides" and lists them: queue is the spec's open children, no ticket cap, spec handoff never fires, plus its own steps. No third reference file.
3. **What is loop-level (shared):** frontier query, slots check with `uptime` and `free -g`, shallow exploration and clumping via in-process Explore subagents writing to `~/.cache/burndown`, dispatch via `implement-dispatch`, waiting on wakes plus vanished-worker polling, merge-line handoff, progress file, cost line, park and the two-park stop, merge tail, re-list.
   **burndown only:** mixed-queue listing by label, ticket cap, spec handoff.
   **implement-spec only:** queue fixed to one spec's children, spec-wide exploration pass first, file-ownership brief, one end-to-end test over the spec, spec-level review at the end.
[...]

PR #926, opened 2026-09-20
The burn loop: frozen candidates, continuous refill, closure freshness and the box check (#893)

### LINKED ISSUE #893: burndown § The loop: dispatch, continuous refill, closure freshness, box check

caneff:
Closes #893

## What changed

`burndown/SKILL.md` § The loop is now the single home for the dispatch loop (#779): where the controller sits and the worktree refusal, the candidate set frozen at a whole-queue exploration with the stuck-run exception, continuous refill recomputed at each landing, the one-hop closure re-resolve at dispatch with the live-workspace exclusion, the hub landing as the only full re-exploration trigger, the box check before every dispatch, and one re-announcement per live unlanded worker on resume.

`burndown/loop.py` is the reader behind it — `seat`, `frontier`/`picks`/`refill`, `hubs`/`hub_landing`, `box_check`/`box_room`, `admit`, `announce`, and a `seat`/`box`/`dispatch`/`hub` CLI. `burndown/references/loop.md` holds why each rule reads that way, with the #781 evidence.

`burndown/loop_test.py` (38 tests) and `burndown/loop-steps.test.sh` guard the behaviour and the prose; the loop restates no `implement-dispatch`, `merge-cleanup` or herdr grammar, and the guard test fails if it starts to.

## Tests run

`bash tests/all.sh` → `56 suites passed`, including `PASS burndown/loop_test.py` (38 passed) and `PASS burndown/loop-steps.test.sh`.

## Decisions made

Every round-1 finding, with its disposition. Dispositions recorded by the verification pass at `~/.cache/agent-reviews/skills/dispositions-893.jsonl`; `S6`–`S9` are the standards report's over-engineering cuts, in the order it lists them.

- `S1` — fixed, `0c4d474` and `0a1e857` (the per-rule reasoning lives in `references/loop.md` once; the verification pass found `seat`/`admit`/`announce` still restating theirs, fixed in `0a1e857`).
- `S2` — fixed, `0c4d474`. The box numbers have one home in `loop.py`'s constants and left `references/loop.md`; § The loop still names 28 and ~24 GB because the ticket's own acceptance criterion names them.
- `S3` — fixed, `0c4d474` (`picks(state, free)` split out, so the CLI reads the frontier once).
- `S4` — fixed, `0c4d474` (the seat CLI test builds its own primary checkout and linked worktree instead of reading whichever seat it runs in).
- `S5` — fixed, `0c4d474` (box boundary test renamed to what it asserts).
- `S6` — fixed, `0c4d474` (`box_check`'s unused `cap=`/`budget_gb=` knobs cut).
- `S7` — disputed: `closure.py` emits both `closure` and `files` in closure mode, and reading `files` under-detects the collision the exclusion rule exists to stop, so the two-key read stays. Precedence is now pinned by a test (`C3`).
- `S8` — fixed, `0c4d474` (reference prose trimmed out of § The loop steps 5–6).
- `S9` — disputed: the frozen candidate set with the stuck-run exception is an acceptance criterion of #893, and § The loop step 2 names `admit`, so it is not an abstraction without a caller.
- `P1` — fixed, `0c4d474` (step 8 registers each dispatched clump with `runfile.py clump`, or resume's announce bucket is always empty).
- `P2` — fixed, `0c4d474` and `0a1e857` (`loop.py dispatch` requires the box readings and refuses rather than picking; `0a1e857` made it weigh every worker the tick would start, after the verification pass found the gate asked about one).
- `P3` — disputed: the ticket body puts the controller on "Any primary checkout's **default branch**", so the detached-HEAD and non-default-branch refusals serve that criterion; `stuck_on` membership is what keeps the frozen-set exception from widening to any ticket filed today; a messenger callable needs a message to send.
- `P4` — fixed, `0c4d474` and `0a1e857` (step 5 names `closure.py`; `0a1e857` replaced the guard needle that step 3 already satisfied).
- `C1` — fixed, `0c4d474` and `0a1e857` (clump files validated field by field; `0a1e857` added the `workspace` an in-flight entry needs).
- `C2` — fixed, `0c4d474` (the fileless-clump refusal fires with nothing in flight too).
- `C3` — fixed, `0c4d474` (a test with `files` a strict subset of `closure` pins the precedence).
- `C4` — fixed, `0c4d474` (a negative free count takes nothing).
- `C5` — fixed, `0c4d474` (a failed re-announce names the workers already reached).
- `C6` — fixed, `0c4d474` (an unset `origin/HEAD` is refused with `git remote set-head origin -a` named).
- `C7` — fixed, `0c4d474` (step 5 says the run file supplies an in-flight clump's tickets and workspace and the re-resolution supplies its closure).

Scope: the diff touches only `burndown/`. The skill stays parked — #899 unparks it after a supervised run — and resolving a herdr agent name to a sendable address stays #923's.

## Round 2 — Codex adversarial pass

- **[high] `paths()` treats a malformed closure as a path set** — fixed, `47702c4`. `set("shared.py")` was a set of six letters that intersected no real path set, so a closure written as a string read as a clump colliding with nobody and dispatched a second worker into a file a live workspace held. A closure or file list is now refused unless it is a non-empty list of non-empty strings, in `paths()` — the chokepoint both the CLI and the in-process callers pass through. Guard mutation-checked.
- **[high] "dispatch ordered before its gates"** — disputed by the controller: step 5 ("At each dispatch, re-resolve the closures") and step 7 ("Box check before every dispatch") are scoped as prerequisites of every dispatch, step 4's included; the numbering is not the execution sequence.
- **[medium] non-idempotent resume announcement when `send` delivers and then raises** — disputed and deferred to #923: a fix needs a stable message id and an acknowledgement, which is that ticket's seam. `announce` keeps taking the messenger as an argument.

## Last reviewed sha

`0c4d474` for the three-axis round; the verification pass read `0c4d474` and its own carry-overs landed in `0a1e857`, which was not re-reviewed by a fresh axis.


caneff:

`codex-companion.mjs adversarial-review --wait --base origin/main`, ticket title + body + comments supplied as context. Verdict: `needs-attention`, three findings. Controller dispositions below; one confirmed by reproduction, two disputed.

### CONFIRMED — [high] Malformed closure values are treated as paths and hide a live-workspace collision (`burndown/loop.py`, `paths`)

Reproduced end to end through the CLI. Two clumps, both naming `shared.py`, differing only in whether the candidate's `closure` is a string or a list:

```
$ loop.py dispatch --candidates <string closure> --in-flight <live> --free 1 --processes 4 --committed-gb 4
dispatch  #1  #1

$ loop.py dispatch --candidates <list closure>   --in-flight <live> --free 1 --processes 4 --committed-gb 4
held      #1  by #2 in /w/2  over shared.py
```

`paths()` calls `set()` on any truthy value, so `"shared.py"` becomes `{'e','a','h','r','d','s'}` and intersects nothing. The loop then dispatches a second worker into a file a live workspace holds — the collision this function exists to prevent.

`read_clumps` does not catch it: `tickets` is validated hard (list, non-empty, positive non-bool ints), then `paths(entry)` is called for its side effect and accepts anything truthy. Its own docstring states that "a hand-built or half-written file is the normal case here," reached "while a controller is recovering" — so the input that triggers this is the input the reader was written for. Same fail-open class that agent-skills#891 closed in `closure.py`, one layer up.

Fix: validate `closure`/`files` as a non-empty list of non-empty strings in `read_clumps` before `paths` is reachable, with a test proving a string, an object, and a list with a non-string member each refuse in one line.

### DISPUTED — [high] The step sequence permits dispatch before closure freshness and box admission (`burndown/SKILL.md`)

Codex read the step numbering as the execution sequence. The steps are scoped per dispatch, not staged after it: step 5 reads "**At each dispatch, re-resolve the closures**" and step 7 reads "**Box check before every dispatch**." Both are stated prerequisites of every dispatch, including step 4's. No change.

### DISPUTED (deferred, not dismissed) — [medium] Resume messaging cannot uphold exactly-once across an ambiguous send failure (`burndown/loop.py`, `announce`)

Real in principle: if `send` delivers and then raises — a transport timeout after remote acceptance — the worker is not recorded as announced and a retry double-sends. Fixing it properly needs a stable message id and an acknowledgement, which is the address-resolution and messaging seam of agent-skills#923, not this ticket's. Filed there rather than built here; `announce` keeps taking the messenger as an argument.

Round 1 of the two-round ceiling. Round 2 carries the confirmed finding only.

Issue #925, opened 2026-09-20
Liveness: a worker spinning on identical no-op tool calls reads as healthy


caneff:
## TL;DR

A worker stuck repeating the same do-nothing command looks exactly like a worker thinking hard. Every check we have says it is fine, and only a human glancing at it notices.

## What to build

A liveness check that flags **N identical consecutive tool calls** from one worker, and reports it the way the stop alert reports a silent stop.

The existing signals all cover the opposite failure — a worker that *stops*. #886 fixed the stop alert's false positives; #900 covers subagent launches read per-turn. Nothing covers a worker that never stops and reports nothing.

Identical-call repetition is the mechanical signature, not `echo ok` specifically: the same tool with byte-identical input, N times in a row, with no other tool call between. Pick N from the measurement below — the observed loop ran 180 consecutive calls, and legitimate repetition (a retry, a poll the lane explicitly sanctions) does not reach double digits with an unchanged input.

Where it lives is the ticket's first decision: the stop hook only fires when a worker stops, which is exactly what this worker never does. It likely needs a check that can fire *during* a turn, or one the controller can run against a transcript on demand. State which and why.

## Evidence

`skills-893`, 2026-09-20, building #893. Chris spotted it from outside; no automated signal did.

- **180 of its last 250 tool calls** were `Bash {'command': 'echo ok'}`, one every ~2 seconds from 11:23:35 to 11:33:28.
- It was polling for a subagent it had launched at 11:30:54 (`Agent`, verification pass) — the poll-instead-of-going-idle antipattern the lane already forbids in words.
- Every liveness signal read healthy throughout: `herdr agent list` said `working`, `interactive_ready: true`, and the transcript was being appended every two seconds, so file mtime looked perfectly alive.
- Its actual work was intact and committed, so the cost was turns and tokens rather than lost work — but the controller had no way to know that without reading 250 tool calls by hand.

The generalisation that makes this worth building: **a worker inside a tool call is unreachable and unobservable.** Time-in-state cannot distinguish this from real work, and mtime cannot either, because the spin writes to the transcript. The repetition itself is the only signal that separates them.

## Acceptance criteria

- [ ] N identical consecutive tool calls from one worker raises an alert naming the worker, the tool, and the repeated input
- [ ] Identical means same tool and byte-identical input, with no other tool call between
- [ ] A worker doing varied work never raises it, however long it runs
- [ ] A sanctioned repeat — a bounded retry, a `Monitor` until-loop — does not raise it
[...]

Issue #896, opened 2026-09-19
merge-tail.md: the collision procedure, the closure feedback loop, and answering before cleanup


caneff:
## Parent

#885

## TL;DR

What to do when two agents' work turns out to overlap after all — which one goes first, how the other catches up, and how to stop the same overlap happening every time.

## What to build

`merge-tail.md`: the controller's procedure when two in-flight branches touch the same files.

**The procedure.** First PR to land wins. The other rebases onto `main`, takes `main`'s side for anything **generated**, and **regenerates** rather than resolving artifact conflicts by hand.

**The feedback loop.** A collision that escaped the include closure (#891) is a defect in the repo's declared include grammar, so the controller files it against the repo. Otherwise the same two files collide on every run and the closure never learns.

**The ordering rule.** The controller answers every outstanding worker question **before** cleanup. Cleanup is the last act, after the worker has been told the outcome.

## Evidence

The procedure is what worked on #781 when `#451` landed and collided with in-flight `#455`: merge one, rebase the other, take `main`'s skyscraper component and layer the branch's own changes, regenerate the link artifacts rather than hand-merging them. No work was lost.

What then went wrong was the harness, not the procedure — the rebase hit a classifier denial that parked the worker for hours. That is cleared now: the hard rules carve out finishing a rebase already under way and a `--force-with-lease` to a worker's own `implement-*` branch, with scoped permission entries beside them. The procedure itself was sound and is worth writing down exactly as run.

The ordering rule comes from a smaller failure with a sharper edge. `#454`'s worker asked for a ruling in good faith. The controller merged, ran `merge-cleanup`, then sent the ruling — and the send failed: `No agent named 'implement-454-12' is reachable`. Cleanup had closed its pane. Nothing was lost materially, but a worker asked a question and never got the answer, and on a longer-lived worker that is a real coordination failure. `implement/SKILL.md` § The merge orders the controller's steps as check, Codex pass, merge, cleanup — and says nothing about answering the worker.

## Acceptance criteria

- [ ] `merge-tail.md` states the first-PR-wins rule and the rebase procedure
- [ ] It states that generated artifacts are regenerated, never hand-merged
- [ ] It requires an escaped collision to be filed against the repo's declared include grammar
- [ ] It requires every outstanding worker question answered before cleanup
- [ ] The loop points at it and does not restate it
- [ ] Guard tests assert each rule is present

## Blocked by

- #891 — "escaped the closure" has no meaning until closures exist
[...]

Issue #778, opened 2026-09-14
Prototype: N workers to one controller over the sessions socket


caneff:
Part of #776

## Question

Does N-to-1 fan-in over the sessions socket wake a parked controller reliably? Start a controller session and three workers via `implement-dispatch` (or three plain sessions in `.claude/worktrees`), have each worker `SendMessage` the controller at different times while the controller is idle, and record what arrives, when, and whether anything is lost or coalesced. Then check the fallback: `herdr agent get` plus the sessions registry when a worker goes quiet. Output: the wait contract for a controller over N workers, ready for the spec, folding in the research ticket's options.

caneff:
## Resolution

**Yes.** N-to-1 fan-in over the sessions socket wakes an idle controller reliably. Four rounds with one throwaway controller and three throwaway workers, all started the way `implement-dispatch` starts a worker:

| Round | Controller | Sends | Result |
|---|---|---|---|
| A | idle | staggered 45 s apart | each message is a new user turn ~1 s after the send; reply 1.5 s later |
| B | blocked in a 150 s tool call | staggered | all three held and shown only when the tool returned (first waited 133 s); none lost, order kept |
| C | idle | burst of 3 in 0.33 s | one turn per message, none merged or dropped, send order kept |
| D | any | worker runs `/exit` | within 8 s `herdr agent get` → `agent_not_found`, registry file and pid gone |

**Wait contract for the spec:**
1. The controller's wait is *going idle*. A worker's `SendMessage` is the next turn.
2. The controller never sits in a long tool call while workers are out; messages are not lost but are unseen until the tool returns. `herdr agent wait` with no timeout, long sleeps, and blocking `TaskOutput` are all off-limits during a run.
3. Concurrent reports serialize, they do not coalesce (at N=3).
4. A vanished worker gives no push signal. The controller finds it by `herdr agent get <name>` plus `~/.claude/sessions/<pid>.json`; both are gone within seconds. `SendMessage --notify_when_idle` is the only push-shaped option and is one-shot per arm.
5. No herdr socket client is needed for fan-in; `events.subscribe` (#777) stays a fallback for a worker that cannot run `SendMessage`.

herdr gotcha: `agent prompt --wait --until idle` times out on a session that ended its turn; herdr reports `done`. Use `--until idle --until done`.

[...]

PR #929, opened 2026-09-20
Liveness, parking and the ruling bar: the loop's wait step gets its policy (#894, #895)

### LINKED ISSUE #894: Liveness: the bounded sweep, the stop alert demoted to a hint, and a worker's declared core count
## Parent

#885

## TL;DR

How the coordinating agent knows its workers are still alive. Right now it relies on an alarm that has cried wolf six times out of six. Replace it with a cheap look at the real state, and have workers say how much of the machine they are using.

## What to build

**Three liveness sources, ranked.**

1. **The wake stays primary.** A worker's `SendMessage` reaches an idle controller as its next turn, about a second after the send (#778). Nothing replaces it.
2. **The stop alert is a hint.** It means "read this pane", never "this worker is stuck". The skill must never instruct a controller to park on a stop alert alone.
3. **The backstop is a bounded sweep.** `herdr agent get` over every live slot, run when the controller wakes for any reason and has nothing else to do. Never a timer, never a blocking call — a controller inside a tool call sees no messages at all until it returns, so it must never sit in one while workers are out. A vanished worker is findable only this way (#778).

**A worker declares its job size.** A worker that launches a parallel job names the job's core count in its report. The controller holds off dispatching into a free slot while a declared heavy job is out.

## Evidence

Across the two trial runs on #781 the stop alert fired **six times and was wrong six times**, with three distinct causes — an outstanding background shell, outstanding Monitor tasks, and a worker that had already reported and was answering a message needing no reply. #886 fixes the hook; this ticket is why the skill must not lean on it even once fixed.

On job size: `#351`'s worker ran `verify.py`, which hard-codes an 8-worker CP-SAT portfolio, at ~793% CPU. Box load hit 25.8 alongside another agent's census **with no dispatch pending**, so no box check could have caught it — the controller's budget is in slots and the real contention is in cores, and nothing bridged the two. The controller's attempted fix made it worse: it ordered the job down to one worker, applying a rule from `AGENTS.md` to a tool with no such knob, which times out at one worker.

## Acceptance criteria

- [ ] The skill names the wake as the primary liveness path
- [ ] The skill states the stop alert is a hint, and never instructs a park on it alone
- [ ] The sweep is bounded, runs only on an idle wake with nothing else to do, and is neither a timer nor a blocking call
- [ ] The sweep finds a worker whose pane is gone and reports it distinctly from an idle one
- [ ] A worker's report names the core count of any parallel job it launched
- [ ] The controller holds dispatch while a declared heavy job is outstanding, and says so in its status line
- [ ] A guard test asserts the skill contains no instruction to park on a stop alert alone

## Blocked by

- #893 — the sweep runs inside the loop's wait step
- #886 — the alert's demotion is only safe to write once the hook stops manufacturing false positives

## Seams under test

- The sweep function over a stub agent list: a working worker, an idle worker, and a vanished pane, asserting a distinct verdict for each and that the call is bounded.
- The report's core-count field, and the loop's hold decision given a report that declares one.

### LINKED ISSUE #895: Parking, escalation, and what a controller must verify before it rules
## Parent

#885

## TL;DR

When a job gets stuck, when it should be handed to a person, and what the coordinating agent has to check before it decides something itself. All three were worked out by hand during the trial and none of them is written down.

## What to build

**Three park causes, and no others.** A clump parks when: a question only Chris can answer arises; a lane-mandated step the harness refuses blocks it; or a worker cannot get its PR to CLEAN. Everything else the controller rules on itself. A park is a comment plus a label swap to `ready-for-human`.

A **parked clump keeps its workspace** and holds its include closure out of the frontier, so parking does not invite a second worker into the same files. **Two consecutive parks with no landing stop the run.**

**Two escalation shapes into `CONTEXT.md`'s Controller entry.** The existing list is a spec-ruling change, a new dependency, an irreversible deletion, a repo Chris does not own. Add:

- *The spec is silent on something the user needs* — not a change to a ruling, because there was no ruling.
- *A lane-mandated step the harness refuses* — the one escalation where the controller **structurally** cannot act, since every route either launders a denial or breaks a hard rule.

**The ruling bar: three clauses.**

1. A ruling about **runtime behaviour** is checked against the thing that ships, not a proxy.
2. A ruling about a **tool's behaviour** cites the tool's source, not its help text.
3. A **Codex finding's recommendation** is evaluated by the controller before it reaches the worker — the controller is not a courier.

And: a controller **may commission a bounded probe** from a worker before escalating, so an abstract question reaches Chris as a table.

## Evidence

Each clause is a ruling that went wrong on #781.

- `#367`'s no-ring header verified green against the headless bundle and shipped; a later slice opened it in the live editor, where `sudoku` documents open 9×9 whatever their width, so 4×4 and 6×6 were broken and the ruling was reversed a slice later.
- `#351`'s "drop to one worker" applied an `AGENTS.md` rule to a tool that hard-codes an 8-worker portfolio and has no such knob. Cost a SIGSTOP and a round trip.
- `#559` was filed from one sentence of `merge-cleanup --help`, which reads as a disjunction where `is_cache` is a conjunction. The worker implemented it, disproved it by dry run, and had to stop and ask. Filed as #875.
- The Codex pass on that same `#559` then recommended **the exact approach the dry run had disproved**, having made the same help-text misreading. Couriering it would have spent the worker's last review round on a regression. What prevented it was the controller having read the source an hour earlier for something unrelated — luck, not process.

Both hard blocks of the burn were the harness-refusal shape: a rebase the lane mandated and the classifier denied, and a `.scratch/` deletion the skill mandated and the classifier denied. In each, obeying the skill and obeying the hard rules both ended at a human.

The bounded probe is the counter-example worth keeping: instead of forwarding three options blind, the controller asked for at most three clue counts, one browser, no givens. It came back in minutes as a table and turned an abstract spec question into a one-message ruling.

## Acceptance criteria

- [ ] Exactly three park causes are named, and the skill says everything else is a controller ruling
- [ ] A parked clump keeps its workspace and holds its closure out of the frontier
- [ ] Two consecutive parks with no landing stop the run
- [ ] `CONTEXT.md`'s Controller entry carries both new escalation shapes
- [ ] The three ruling clauses are stated in the skill
- [ ] The bounded probe is stated as something a controller may commission before escalating
- [ ] Guard tests assert each clause is present and that no park cause beyond the three is listed

## Blocked by

- #893 — parking and escalation are steps of the loop

## Seams under test

- The skill's park, escalation and ruling sections, guard-tested for each named cause and clause, and for the absence of any fourth park cause.
- `CONTEXT.md`'s Controller entry — the seam is its stated escalation list, read by every controller.

caneff:
Closes #894

Closes #895

## What changed

`burndown/SKILL.md` gains § Liveness (the wake primary, the stop alert demoted to a hint, a bounded `herdr` sweep as the backstop, and a worker's declared core count charged against the free slots), § Parking and escalation (three park causes and no fourth, a parked clump keeping its workspace and its closure off the frontier, two consecutive parks stopping the run) and § Before a controller rules (the three ruling clauses). `burndown/loop.py` gains the readers behind them — `sweep`, `core_room`/`render_cores`, `parse_declared` — plus `loop.py sweep` and `dispatch --declared`; `CONTEXT.md`'s Controller entry gains the two escalation shapes. Evidence lives in `burndown/references/liveness.md` and `references/parking.md`, guards in `burndown/liveness.test.sh` and `burndown/parking.test.sh`.

§ The loop step 8 also carries the `references/merge-tail.md` pointer line the controller handed over verbatim on #896's behalf, since one writer per file — that is where #896's AC5 landed.

## Tests run

`bash tests/all.sh` → `59 suites passed` (exit 0), at the tree of the reported sha, rebased onto `f3ccd00`. `python3 burndown/loop_test.py` → `76 passed` (#896's 12 included); `python3 burndown/runfile_test.py` → `49 passed`.

## Decisions made

- **`implement/SKILL.md` untouched, by controller ruling.** #894's AC "a worker's report names the core count of any parallel job it launched" lands in two halves: the controller-side policy and reader here, and the "PR up" report line plus its needle in `implement/pr-up-report-shape.test.sh`, written by `skills-896`, which owns that file this run. The peer's contract, which this skill's prose was tightened to agree with: the report names each parallel job and its core count, and says so explicitly when there were none.
- **"Heavy" is a charge, not a threshold.** Every declared job is charged: the cores past the job's own slot come off the free slots, so a 2-core job holds one more and an 8-core job holds seven. The #351 case the AC was written from produces a full hold.
- **#895's "two consecutive parks" is prose, not a Python reader**, per the ticket's own Seams under test, which name the skill's sections and `CONTEXT.md`'s Controller entry.
- **#925 is a stated blind spot, not solved here.** The sweep answers whether a pane exists and what herdr says it is doing, never whether work is progressing.

### Round-1 findings, each with its disposition

Ids are the reviewers' sidecars' (`~/.cache/agent-reviews/skills/findings-{standards,spec,correctness}-894.jsonl`); dispositions in `dispositions-894.jsonl`.

- **S1** the skill restated the reference's evidence — fixed, `fe23a7f`
- **S2** the stop-alert guard defeatable by a stray negation — fixed, `fe23a7f`
- **S3** `read_workers` duplicated `read_clumps` — fixed, `fe23a7f`
- **S4** `core_room` mixed computation and rendering — fixed, `fe23a7f`
- **S5** `CONTEXT.md`'s Controller entry a run-on — fixed, `fe23a7f`
- **P1** "heavy" undefined; charge model vs a binary hold — fixed, `fe23a7f` (the charge model is now stated in § Liveness)
- **P2** the guard's exclusions string-level, not sense-level — fixed, `fe23a7f`
- **C1** a missing `herdr` read as every worker dead — fixed, `fe23a7f`
- **C2** the cores hold exited 1 with the box's reason — fixed, `fe23a7f`
- **C3** no time bound on the probe — fixed, `fe23a7f`; its witness in `b361b01`
- **C4** a clump declared twice, last-wins — fixed, `fe23a7f`
- **C5** hollow witness on #894's load-bearing criterion — fixed, `fe23a7f`; the review's own injected sentence is now a failing case
- **C6** one nameless clump discarded gathered verdicts — fixed, `fe23a7f`
- **C7** new refusal code shipping unexercised — fixed, `fe23a7f`
- **C8** the merge-tail pointer not asked for by #894/#895 — `disputed: the controller handed this worker that line verbatim on #896's behalf, citing #896's own AC5 and one writer per file; the verification pass ruled the reason holds`. Its second half, the parked-spec paragraph left dangling under the new sections, was fixed in `fe23a7f`.
- Over-engineering, `cores_of` validating at both `parse_declared` and `core_room` — `disputed: two boundaries, and core_room is called directly with values parse_declared cannot produce; the verification pass ruled the reason holds`. The other cuts (the verdict map, the unread `unreachable` bucket, the `charged`/"nothing declared" branch, the `read_workers` copy) were fixed in `fe23a7f`.

Verification pass: 15/15 round-1 ids disposed, no orphans, no new test hollow. Its three residuals: **V1** (the timeout had no witness) and **V3** (the held-clump lines stopped printing behind the cores early return) fixed in `b361b01`; **V2** — a sentence whose `never` is not the prohibition ("The stop alert never lies, so hold the slot") still passes the guard — `disputed: the guard is a backstop against the instruction the trial actually produced, not a parser for prose written to defeat it`.

### Codex round 2 — both findings fixed

Comment: https://github.com/caneff/agent-skills/pull/929#issuecomment-5750011113. Both confirmed by the controller from branch source, both fixed in `5ec81f4`.

- **A missing declaration charged as zero, and no declaration surviving a resume.** `--declared` is gone. A clump's parallel job is state in the run file — `runfile.py job <run-id> --clump <n> --cores <k> | --none | --done`, held as `running`/`none`/`done` — and `loop.py dispatch` derives the charge from that record, so it survives the restart an argv could not. A live clump with **no record** is refused by name rather than charged zero: the reader now obeys the rule the skill states one layer up. A #892-era run file with no `job` field still loads, as `null`.
- **The sweep blocking for 10s per live worker.** One deadline now covers the whole sweep; each probe is handed what is left of it, and a slot the deadline did not reach is `unswept` — read on the next wake, never confused with a pane that answered. The worst case is a constant, not N × the timeout.

### Rebased onto #896

`burndown/loop.py` and `burndown/loop_test.py` collided with #930 (#896), which landed first. Rebased per `burndown/references/merge-tail.md` — this branch is that procedure's first user. Every conflict was additive (its landing tail, this branch's sweep and core charge, at the same insertion points), so all three were hand-merges keeping both sides; nothing is generated in this repo, so nothing was regenerated, and paths were staged by name. `landing_steps`, `cleanup_ready` and `questions` are present and contribute nothing to this branch's diff.

## Last reviewed sha

`7ddbe45` (`fe23a7f` before the rebase) — the verification pass read that tree. Two commits landed after it and were not re-reviewed: `32e5db9`, the witness for the probe's time bound and the held-clump lines; and `5ec81f4`, Codex round 2's two fixes.


caneff:

`codex-companion.mjs adversarial-review --wait --base origin/main`, both tickets' title + body + comments supplied as context (115 lines). Verdict: `needs-attention`, two findings. **Both confirmed by the controller from the branch source.** Round 2 carries both.

### CONFIRMED — [high] A missing declaration is charged as zero cores, and no declaration survives a resume (`burndown/loop.py`, `main`/`core_room`)

`--declared` is `default=""`, so `parse_declared("")` yields `{}` and `core_room` charges nothing: a dispatch where the controller did not manually re-supply every declaration treats every outstanding worker as consuming exactly its own slot. That is the same collapse the skill forbids one layer up — § Liveness says a worker that launched none **says so explicitly**: *silence is not zero*, with a guard needle on the phrase — while the reader's own default reads silence as zero.

The file argues against itself, twelve lines above the flag:

```
    # Required, not optional: the box check runs before *every* dispatch, and
    # an optional flag is the step a controller forgets.
    dispatch.add_argument("--processes", type=int, required=True)
    dispatch.add_argument("--committed-gb", type=float, required=True)
```

`--declared` is the one core-budget input that is optional, and it is optional for the exact reason the comment gives for making the others required.

Second half, and the more expensive one: the declaration lives nowhere but that one argv. It is absent from the run-file schema, so a controller that restarts mid-run cannot recover an in-flight 8-core job's hold and will dispatch straight into the contention #351 produced — a run with a free slot dispatching into a box at 25.8 load, which is the incident this criterion exists to prevent.

Fix: persist per-clump job state in the run file — declared count, explicit `none`, and completion — and have `dispatch` derive holds from that state rather than from argv. A live, unlanded clump with no current declaration is not zero: hold conservatively or refuse by name. Test the resume path and the no-declaration path separately.

### CONFIRMED — [high] The sweep blocks the controller for up to 10s per live worker (`burndown/loop.py`, `sweep` / `herdr_get`)

`sweep` iterates `workers` serially and each `get(agent)` is a synchronous `subprocess.run(..., timeout=HERDR_TIMEOUT)` with `HERDR_TIMEOUT = 10.0`. The per-probe bound is real, but it composes: N hung panes hold the controller inside one tool call for up to 10N seconds. At the wave size this map has been running — three to five live workers — that is 30 to 50 seconds.

The docstring states the invariant it breaks: "It is the backstop under the wake, not a substitute for it, and it is never a timer: **a controller inside a tool call hears no worker at all (#778)**." A sweep that can sit 50 seconds inside one tool call is not a timer, but it is the same deafness by another route, and #894's own criterion says the sweep is "neither a timer nor a blocking call".

Fix: one deadline for the whole sweep rather than one per probe — probes that have not answered when it expires are reported as their own verdict and collected on a later wake. The worst case must be bounded by a constant, not by the number of live workers. Guard-test multiple hung probes, which is the case the current tests do not construct.

### Controller notes

Not findings, recorded so the round is not relitigated:

- The step-8 `merge-tail.md` pointer that the correctness axis flagged as unasked-for scope (C8) was ordered by the controller to resolve a cross-ticket collision with #896, whose AC5 it satisfies. Disputed correctly by the worker, upheld by the verification pass, and quoted in this PR body. It stands.
- "Heavy" as a **charge** (cores past the worker's own slot) rather than a threshold is accepted. A threshold needs a number that is wrong for some job; the charge uses the arithmetic the box cap already uses, and #351's 8-core case still produces a full hold. The spec axis reached this independently.
- The body file for this pass was built with `--json title,body,comments` for both tickets and checked non-empty before the run.

Issue #928, opened 2026-09-20
closure.py mis-reads a declaration whose colon sits inside the emphasis


caneff:
`burndown/closure.py`'s `_KEY` regex expects the colon outside the emphasis, so a repo that writes `- **Directive:** \`#include <path>\`` — as common in the wild as `- **Directive**: ...` — yields a directive whose value begins with `**`. It fails silently: the resolver reports a declaration it cannot match against any file, and the run clumps as though the repo declared a directive that regenerates nothing.

Evidence: `python3 -c "import sys; sys.path.insert(0,'burndown'); import closure; print(closure.declaration_section('## Include closure\n- **Directive:** x\n'))"` shows the `**` on the value. Found while building #897's `## End-to-end seam` parser, which accepts both forms — `implement-spec/closing_ticket.py`'s `_KEY`, with a lookahead so a value that is itself bold keeps its markers. The fix is to bring the same clause to `closure.py` (and, once both agree, to share one key parser rather than two).

Filed from: the correctness review on #897 (finding C2), whose sibling in `closure.py` was out of that ticket's scope.

## Blocked by

- #897

Issue #927, opened 2026-09-20
agent-skills declares its own end-to-end seam in AGENTS.md


caneff:
#897 specifies and tests the `## End-to-end seam` grammar that a closing ticket reads — the repo's end-to-end check plus what that check is blind to — but proves it against fixture repos only. This repo has never declared its own seam, so the first real `implement-spec` run here reads no declaration and falls back to whatever the exploration pass infers.

What to do: add an `## End-to-end seam` section to this repo's `AGENTS.md` in the grammar #897 lands, naming `bash tests/all.sh` as the seam and stating what it cannot see. The blind spots are known and measured, not hypothetical — the gate is 56-59 shell and Python suites with no rendered surface in it, so anything a human reads rather than a test asserts is outside it: a skill's prose being followed by a model, a `SKILL.md` instruction that is present but ambiguous, and the harness behaviours around it (a hook denying a step, a worker spinning on no-op tool calls, a pane closing before a question is answered — see #925).

Where: `AGENTS.md` at the repo root, section `## End-to-end seam`, grammar and reader specified by #897 in `implement-spec/closing_ticket.py` and `implement-spec/references/closing-ticket.md`.

Why it was not done in #897: `AGENTS.md` was contended by three in-flight workers during map #776's last wave, and no acceptance criterion asked for it. The worker flagged it rather than widening its own diff.

Filed from conversation — `skills-897`'s report to the controller during map #776.

Issue #890, opened 2026-09-19
The frontier reader: native dependencies, a stated Blocked-by grammar, and unresolved as its own answer


caneff:
## Parent

#885

## TL;DR

To pick the next piece of work, an agent needs to know which tickets are ready and which are still waiting on something else. On some trackers that relationship is a real, queryable link. On others it is a sentence someone typed in the ticket body — and on some tickets, nobody typed it at all.

## What to build

A **frontier** reader: given a repo and a label, return the tickets that are open, unclaimed, and not waiting on anything.

Three sources, in order:

1. **Native dependencies** where the tracker has them. On GitHub that is `issue_dependencies_summary.blocked_by`, which counts open blockers only and is the live gate. This is the canonical path.
2. **A stated `## Blocked by` grammar** as the fallback, for a repo whose tickets carry the relationship as prose. Specify the grammar the reader parses — the section heading and the `#NNN` reference form — rather than parsing whatever it meets.
3. **Nothing at all.** A ticket with no native dependency data *and* no `## Blocked by` section is **unresolved**, not unblocked. It is reported as unresolved and never dispatched on the assumption that silence means clear.

Then close the loop at the source: `/to-tickets` emits native dependency edges where available and the `## Blocked by` section otherwise, so the fallback reader gets better inputs over time.

## Evidence

From the burn on #781, on `sudokumaker-custom-constraints`: the repo had no native blocking links, edges lived in a `## Blocked by` markdown section, and **8 of 21 `ready-for-agent` tickets had no such section at all**. Computing one wave meant fetching all 21 bodies and regex-parsing them by hand. That is acceptable for a supervised run and wrong as a skill step, because a skill cannot say "take the frontier" and mean a query that does not exist.

Reading a missing section as "unblocked" is the expensive failure: it dispatches a worker onto a ticket whose prerequisite is still open, and the cost is a whole build.

## Acceptance criteria

- [ ] Native dependencies are used where the tracker provides them
- [ ] The `## Blocked by` grammar is specified in the skill and parsed by the fallback reader
- [ ] A ticket with neither source reports as **unresolved**, distinct from both blocked and unblocked
- [ ] A ticket whose blockers are all closed reads as unblocked
- [ ] A claimed ticket (assignee, or `in-progress`) is off the frontier
- [ ] `/to-tickets` emits native edges where available and the `## Blocked by` section otherwise
- [ ] Every case above has a test that fails before the change

## Blocked by

None — can start immediately.

## Seams under test

[...]

PR #931, opened 2026-09-20
implement-spec and spec-handoff.md: the nested run, the contradiction check, the closing ticket's seam (#897)

### LINKED ISSUE #897: implement-spec and spec-handoff.md: the nested run, the contradiction check, the closing ticket's seam
## Parent

#885

## TL;DR

Rebuild the skill that takes a whole plan and drives its pieces through several agents. It runs inside a larger run, and it has three specific habits that wasted people's time last go: it flags things as contradictions when they are just not built yet, and it asks a closing agent to write a test without saying where to put it.

## What to build

`implement-spec` and `spec-handoff.md`, as **policy over `burndown § The loop`** with overrides, restating none of it (#779).

**The nesting.** `implement-spec` is a worker to the burn that dispatched it and a controller to its own slices. Its slot budget is **debited from that burn**, so nesting does not double-spend the box.

**The contradiction check.** The exploration pass checks the spec's decisions against the code. It must compare a contradicted decision against **the spec's own ticket list first**: a decision the code does not implement yet, where a ticket in this spec builds it, is "not yet built" and goes in the exploration summary as a line. Only "the code does this, **differently**" reaches Chris as a contradiction.

**The closing ticket.** It names the repo's end-to-end seam **and what that seam is blind to**, from `AGENTS.md` or the exploration pass. Where the spec has a user-visible surface the seam cannot reach, it says so, and its acceptance includes one open of the real thing.

**The spec-level review** is handed the **list of merge shas** from the run file, not a git range.

## Evidence

All from the spec run on #781.

The contradiction check flagged two decisions on `#366` as CONTRADICTED — `RULES_PREFIX` hardcoded in `build_doc`, and no no-ring path in `framebuild.py`. Both described exactly the gap that ticket `#367`, inside the same spec, existed to close. That is not a contradiction; it is the spec correctly describing a future state. Every spec with a prefactor slice generates this noise, into the one channel that reaches Chris.

The closing ticket said "write one end-to-end test over the whole spec's acceptance criteria" and named no seam, so the closing worker stopped and asked. The seam that existed was the headless solver bundle — which had already diverged from the live editor **inside that same spec**, when the no-ring header verified green headless and broke 4×4 and 6×6 in the real app. So naming the seam is necessary and not sufficient: the worker needs its blind spot too.

The closing ticket's "review the whole landed range since the spec's first slice" is not a clean git range on a shared `main`. `a866bf3..origin/main` held the spec's three squash commits and **~17 unrelated commits from other sessions**, and `/multi-axis-code-review` takes one fixed point.

## Acceptance criteria

- [ ] `implement-spec` points at `burndown § The loop` and restates none of it
- [ ] Its slot budget is debited from the dispatching burn
- [ ] The contradiction check compares against the spec's own ticket list before escalating
- [ ] "Not yet built" appears as a summary line, never as a ruling for Chris
- [ ] Only "the code does this differently" escalates
- [ ] The closing ticket names the e2e seam and its blind spot
- [ ] Where a user-visible surface is beyond the seam, the closing ticket says so and its acceptance includes one open of the real thing
- [ ] The spec-level review receives a sha list, not a range
- [ ] `spec-handoff.md` is a pointer, not a restatement

## Blocked by

- #893 — implement-spec is the loop with overrides
- #888 — the spec run's merge tail inherits the Codex step, which must not mandate a run whose input cannot change

## Seams under test

- The contradiction check over a fixture spec: a decision built by a ticket in the spec, a decision the code implements differently, and a decision the code does not mention. Assert which of the three escalates.
- The closing ticket's generated body — the seam is what it names, asserted against a fixture repo declaring an e2e seam and a blind spot.

caneff:
Closes #897

## What changed

`implement-spec` comes back as **policy over `burndown/SKILL.md` § The loop**, restating none of it (#779) and overriding only the four things a spec run does differently: the nesting, the exploration pass, the closing ticket, the spec-level review.

Two readers carry the mechanical halves: `implement-spec/contradictions.py` classifies the exploration pass's decisions against the spec's own ticket list, and `implement-spec/closing_ticket.py` generates the closing ticket's body from the repo's declared end-to-end seam. `burndown/references/spec-handoff.md` becomes a pointer.

`burndown/SKILL.md` is untouched — it is another worker's file in this wave, and the pointer is honest as it stands.

## Tests run

`bash tests/all.sh` → `59 suites passed` (re-run at `1cf6ff6`) (56 on `main`; the three new ones are `implement-spec/contradictions_test.py`, `implement-spec/closing_ticket_test.py`, `implement-spec/nested-run.test.sh`).

## Decisions made

**Round-1 findings, every one with its disposition** (ids from the review sidecars; the verification pass recorded the same set in `dispositions-897.jsonl`):

- **S1** — fixed in `3f733c7`. The seam declaration is read through #890's fence reader, imported from `burndown/frontier.py`: a grammar shown in a fence declares nothing.
- **S2** — filed as #928. The fence reader is now genuinely shared; the small `- **Key**: value` regex stays local and now diverges deliberately (it accepts a colon inside the emphasis, which `closure.py` does not — a real bug there). Convergence on one parser belongs in that ticket.
- **S3** — fixed in `3f733c7` (tautological `or`, both arms the same literal).
- **S4** — fixed in `3f733c7`; the CLI assertion is keyed to the decision id, not to the accident of a key named `decision`.
- **S5** — `disputed: ~/.agents/skills/… is this repo's cross-skill link convention (burndown/SKILL.md uses it) and tests/check-section-references.py::resolve special-cases that prefix. The link resolves under the repo's own checker.` Upheld by the verification pass.
- **S6** — `disputed: check_not_in matches with a word-boundary regex, so no substring misfire; a future sentence that restates a loop rule tripping the gate is the anti-drift guard working, and the FAIL names the needle.` Upheld by the verification pass.
- **P1** — fixed in `3f733c7`. The surfaces question must be answered: `None` is refused, `[]` says the seam reaches everything, and the CLI's `--surface`/`--no-surface` are mutually exclusive and required.
- **P2** — fixed in `3f733c7`; the duplicated slot-budget bullet is gone from the handoff and points at the nesting instead.
- **P3** — fixed in `3f733c7`; the generated criterion says the test lives where the seam runs it.
- **C1** — fixed in `3f733c7`. `references/exploration.md` now draws the line between `absent` and `differs`, with the worked `RULES_PREFIX` example on both sides and the tiebreak: where both readings fit, read `absent`, because Chris's channel is the one that has to stay clean.
- **C2** — fixed in `3f733c7` (`- **Seam:** x`, colon inside the emphasis, no longer yields a value beginning with `**`).
- **C3** — fixed in `3f733c7`; a root that is not a directory is refused as such rather than as a missing declaration, and the CLI's positionals are typed.
- **C4** — fixed in `3f733c7`; a non-UTF-8 `AGENTS.md` is decoded leniently, not crashed on.
- **C5** — fixed in `3f733c7`; the empty-sha refusal and the ticket-number validation get their witnesses.
- **C6** — fixed in `3f733c7`; `read off the run file` replaces a needle the loop-pointer paragraph satisfied, and the tautological handoff needle is gone.

Over-engineering items from the same round: `stdlib: argparse` fixed in `3f733c7`; `delete: Result.verdict` `disputed: verdict is the documented three-verdict seam that references/exploration.md tabulates and three tests assert on` (upheld); `yagni: private parser` is S2 above.

**Codex round-2 findings (PR comment 5750035541), all three fixed in `1cf6ff6`:**

- **FIX 1** — the sha-list review instruction had no executable path. `/multi-axis-code-review` pins one fixed point and reads `<fixed point>...HEAD`, so "review these shas" was a procedure nothing could carry out and a worker would have invented the range the criterion exists to prevent. The generated ticket now carries the procedure that builds the comparison: a detached worktree at the first sha, the rest cherry-picked on in landing order, the review against `<first>~1`, the worktree removed after — plus the per-sha fallback written out for a conflicting cherry-pick. A native sha-list mode for the review skill is #932 (already open; my duplicate #935 is closed against it).
- **FIX 2** — exploration no longer overrides a declared seam. The `## End-to-end seam` declaration is authoritative, the pass fills only what the declaration omits, and both present and differing refuses with a diagnostic naming each source and its value. The reference text and the docstring now say what the code does.
- **FIX 3** — one invalid `built_by` no longer suppresses every real contradiction. An entry the reader cannot classify becomes an `error` line against that decision and the rest are still read, so a `differs` behind it still reaches Chris; the CLI prints the errors on stderr and exits non-zero. A malformed **ticket list** still stops everything, since that is what every defence is checked against. Mixed fixture: `test_one_malformed_decision_does_not_hide_a_real_contradiction`.

**Other decisions:**

- Both files keep their **Parked** banner. Unparking `burndown` and `implement-spec` is #899's whole job, under one supervised run after every sibling lands; a worker unparking its own skill makes that finish line unverifiable.
- This repo's own `## End-to-end seam` block in `AGENTS.md` is **not** in this diff: `AGENTS.md` is contended in this wave and no acceptance criterion asks for it. Filed as #927, blocked by this ticket.
- `closing_ticket.py` imports `unfenced` from `burndown/frontier.py` by path. That coupling is already real — `implement-spec` is policy over `burndown` and does not run without it — and the alternative is a second fence parser, which is what S1 was.
- The `Closes #897` trailer sits on `a0d7d73` rather than on the tip: `3f733c7` is the sha the review dispositions join against, so amending it would break that join. The PR body carries the closing line, which is what registers `closingIssuesReferences`.

## Last reviewed sha

`1cf6ff6` — the Codex round-2 pass ran against `3f733c7`, and `1cf6ff6` is its three fixes. No commit follows it, and nothing after this point would have been re-reviewed.


caneff:

`codex-companion.mjs adversarial-review --wait --base origin/main`, ticket title + body + comments as context. Verdict: `needs-attention`, three findings. **All three confirmed by the controller from the branch source.**

### CONFIRMED — [high] The SHA-list review instruction has no executable path (`implement-spec/closing_ticket.py`)

The generated closing ticket tells a worker to run `/multi-axis-code-review` over a list of merge SHAs. That skill takes **one** fixed point and builds `git diff <fixed-point>...HEAD` — its own SKILL.md, § Pin the fixed point and § capture the diff command. It cannot consume disjoint landed commits. So the ticket instructs a procedure no skill can execute, and the worker is left to invent a range: the shared-`main` failure this criterion exists to prevent.

This is the fourth instance of one defect class in this trial — a stated fallback with no mechanism behind it (agent-skills#888's incidentally-justified rule, agent-skills#877's guard that cannot tell a citation from a use, agent-skills#911's `--blocked-by` fallback that no code path reached). A rule is not shipped until something can carry it out.

Fix, scoped to this ticket: the closing ticket emits a **concrete executable procedure** over the sha list — per-sha diffs, or an explicitly constructed comparison — rather than naming a skill that cannot take its input. A sha-list input mode for `/multi-axis-code-review` is a separate ticket against that skill and is not to be built here.

### CONFIRMED — [high] Exploration overrides a declared seam (`implement-spec/closing_ticket.py`, `seam_of`)

```python
seam = (seam or found.get("seam") or "").strip()
blind_to = (blind_to or found.get("blind to") or "").strip()
```

The exploration pass's arguments take precedence over the repo's `AGENTS.md` declaration, so a stale or mistaken exploration result silently replaces the canonical seam and blind spot. `references/closing-ticket.md` states the opposite — exploration supplies seam data **when a repo declares nothing**. The docstring states the code's behaviour, so the doc and the code disagree inside one PR.

Ruling: the declaration wins. A declaration that an inferred value can silently override is not a declaration, and agent-skills#927 exists precisely to make this repo's own seam canonical. Use a complete `AGENTS.md` declaration as authoritative; use exploration only where the declaration is absent; when both are present and differ, **refuse with a diagnostic naming both** rather than selecting one. Align the reference text and the docstring with whichever half changes.

### CONFIRMED — [medium] One invalid `built_by` suppresses every real contradiction (`implement-spec/contradictions.py`, `check`)

```python
return [_verdict(_validated(d, numbers)) for d in decisions]
```

`_validated` raises on the first bad decision, so nothing is emitted for any decision. An unrelated stale `built_by` naming a ticket outside the spec therefore hides a later `found: differs` — the one verdict that reaches Chris. The intent is fail-closed and the effect is a blind spot: the check exists to surface contradictions and one malformed neighbour silences all of them.

Fix: report invalid entries as explicit exploration errors **and** continue classifying independent decisions, so a real contradiction still reaches Chris alongside the error. Add a mixed fixture — one malformed `built_by`, one genuine `differs` — proving the contradiction survives.

### Controller notes

- The hard finding from the worker's own round 1 (the seam parser reading a fenced example as a declaration) was real and is fixed by importing agent-skills#890's `unfenced` read-only. Good reuse; no edit to `burndown/frontier.py`.
- agent-skills#928 filed against `burndown/closure.py`'s key regex, with a blocked-by edge, rather than diverging silently. Correct.
- `Closes #897` sits on `a0d7d73` rather than the tip so the dispositions sidecar's join on `3f733c7` survives. Accepted — GitHub registered the closing reference, which is what `merge-cleanup` reads.
- Body file built with `--json title,body,comments` and checked non-empty before the run.

PR #930, opened 2026-09-20
The merge tail: the collision procedure, the closure feedback loop, and answering before cleanup (#896)

caneff:
Closes #896

## What changed

`burndown/references/merge-tail.md` comes back from its parked stub as the controller's procedure for two in-flight branches that turn out to touch the same files: first PR to land wins, the other rebases, generated artifacts are regenerated rather than hand-merged, and a collision that escaped the include closure is filed against the repo's declared grammar so the closure learns.

Its ordering rule has a reader behind it — `loop.landing_steps` returns a landing's tail as answer, then merge, then cleanup, `loop.cleanup_ready` refuses cleanup while a worker question is outstanding, and `python3 burndown/loop.py landing` is the gate a controller runs before `merge-cleanup`.

`implement/SKILL.md` § The merge step 5 carries the same rule at the step that runs cleanup, and § The PR gains #894's `Parallel jobs:` field (see Decisions made).

## Tests run

`bash tests/all.sh` → `57 suites passed`. New: `burndown/merge-tail.test.sh` (prose guard, rule by rule), and 8 cases in `burndown/loop_test.py` (`52 passed`).

## Decisions made

**Acceptance criterion 5** — "The loop points at it and does not restate it" — is satisfied by a pointer line in `burndown/SKILL.md` § The loop step 8, written by the #894 + #895 clump's PR at the controller's direction, not by this PR: that file has one writer this wave and it is not this workspace. The line handed to that worker, verbatim:

> When two in-flight branches turn out to touch the same files, the collision procedure, the closure defect it implies, and the rule that every outstanding worker question is answered **before** cleanup: [`references/merge-tail.md`](references/merge-tail.md).

**`implement/SKILL.md` § The PR's `Parallel jobs:` field** is #894's criterion, not #896's, landed here at the controller's direction because this file has one writer this wave. Its controller half (`loop.py dispatch --declared`) is `skills-894`'s and unlanded, so the field is collected and not yet consumed.

**`implement/SKILL.md` § The merge step 5** is in scope on the controller's ruling: #896's own evidence is that this step ordered check, Codex pass, merge, cleanup and said nothing about answering the worker.

**No CLI subcommand for the answer itself.** `landing` reports what is owed and gates cleanup; sending is `SendMessage`, which the controller calls — the same reason `announce` has no subcommand.

### Round-1 findings, every one with its disposition

[...]

Issue #776, opened 2026-09-14
Multi-worker lane on herdr: implement-spec and burndown rebuilt


caneff:
## Destination

A spec, ready for `/to-tickets`, that rebuilds `implement-spec` and `burndown` (with `spec-handoff.md`) on the herdr lane, written from one supervised run. The skills come back as policy over the `/implement` lane's own primitives (`implement-dispatch`, the sessions socket, `herdr agent`, `merge-cleanup`); nothing Orca-shaped survives.

## Notes

**Domain.** A run is one controller session on the primary checkout's default branch driving N workers, each in its own workspace via `implement-dispatch`. Terms as `~/.agents/skills/CONTEXT.md` defines them: Dispatcher, Worker, Workspace, Brief, Controller, Tier, Front end. The old lane is at `git show 7c7eb30:burndown/SKILL.md` and `git show 7c7eb30:implement-spec/SKILL.md`; the parking commit is #752. The prior multi-worker map is [Lane 3: one spec as a supervised Orca task DAG](https://github.com/caneff/agent-skills/issues/492); the `/implement` lane spec is [#709](https://github.com/caneff/agent-skills/issues/709).

**Skills.** `/grilling` and `/domain-modeling` on every decision ticket. Ponytail applies: a skill that names a policy and defers to `implement-dispatch`, `herdr agent` or `merge-cleanup` is fine; one that restates their grammar has drifted.

**Settled while charting (2026-09-13) — do not relitigate:**

- **The lane is proven.** `~/orca/workspaces` is gone and the finish-line ticket #720 is closed; nothing gates this rebuild.
- **Both skills come back, as two skills.** `implement-spec` keeps the spec-level steps (exploration task first, file-ownership brief, one end-to-end test over the whole spec, spec-level review at the end). `burndown` still hands a spec-labelled sub-issue's whole spec to `implement-spec`. Which file owns the shared dispatch loop is a ticket.
- **A run reuses `/implement` whole.** Each worker follows `/implement` end to end at its tier (self-review, PR, stays alive). The controller claims, dispatches, waits, and hands Chris merge lines. Controller-owned review and "builders run no reviews" are retired.
- **One workspace per worker, one PR per slice.** The shared spec workspace and one-PR-per-spec are retired. Slice order is carried by the tracker's blocking edges.
- **The tracker is the DAG.** Frontier = open, unblocked, unclaimed children, the same query wayfinder uses. No local run file.
[...]

PR #940, opened 2026-09-20
The frontier reader answers in four buckets: a spec parent is its own answer (#910)

### LINKED ISSUE #910: The frontier reader drops a spec parent instead of reporting it unresolved
## Parent

#885

## TL;DR

The tool that picks the next piece of work reports the spec document itself as something it cannot classify, every single time it runs. It is not work; it should not be in the answer at all.

## What to build

#890's reader returns `{unblocked, blocked, unresolved}`. A `spec`-labelled issue has no blockers of its own and states none, so it lands in `unresolved` on every run.

`unresolved` means "this ticket's blocking state could not be determined, so a human must look". A spec parent is not that. It is a ticket that is never dispatchable by design — `implement-dispatch` already refuses a `spec`-labelled issue outright — so reporting it as needing a human look is a false signal that costs a controller a decision every burn.

The reader should drop it before classification, the same way it drops a claimed ticket, and for the same reason: it is off the frontier by its own nature, not by a blocking relationship.

Check whether any other label in this repo marks a ticket non-dispatchable in the same way, and if so treat them together rather than special-casing `spec` alone.

## Evidence

Measured on `caneff/agent-skills` by `skills-890` after #890's fix landed on its branch: `ready-for-agent` reads 6 unblocked, 7 blocked, 8 unresolved. #885 is one of the eight. All twelve of #885's own children classify correctly — the board's spec work is fine, and this is the parent showing up in its own frontier.

`unresolved` exists to catch silence that would otherwise dispatch a worker onto a blocked ticket. Padding it with an item that can never be dispatched trains a controller to skim the bucket, which is the failure the bucket was created to prevent.

## Acceptance criteria

- [ ] A `spec`-labelled issue is absent from all three buckets, not reported as `unresolved`
- [ ] Any other label that makes a ticket non-dispatchable is handled the same way, or the ticket states that `spec` is the only one
- [ ] The measured `unresolved` count on this repo drops by exactly the number of spec issues in the queue
- [ ] A test fixture covers a `spec` ticket with no `Blocked by` section and asserts it is dropped, not classified

## Blocked by

- #890 (it builds the reader this changes)

## Seams under test

- The frontier function's filter step: `(repo, label) -> {unblocked, blocked, unresolved}` against a fixture queue containing a spec parent, a claimed ticket, a well-formed child and a silent one.

Filed from a measurement by `skills-890` during #890's build, ruled by controller `skills-dc`, 2026-09-20.

caneff:
Closes #910

## What changed

`burndown/frontier.py` gains a fourth bucket: a `spec`-labelled parent lands in `spec`, not `unresolved`, and its entry carries the verb that dispatches it — `implement-dispatch --spec <n> --slots <k>` — so a controller can act on the line without opening another document.

A `needs-info` ticket now drops off the frontier entirely, the way a claimed one does: `implement-dispatch` refuses it on the label alone, so it is not work this reader may offer.

`blocked` outranks `spec` on one entry. The frontier checks prerequisites before it offers a route, so a spec parent with an open blocker reads `blocked`.

## Tests run

`python3 burndown/frontier_test.py` → `60 passed`. Every test file in the repo (51 files, `*_test.py` and `*.test.sh`) → exit 0.

## Decisions made

**AC1 and AC3 are the amended ones.** The ticket body says drop a spec parent from all three buckets; Chris's ruling comment (2026-09-20) supersedes that with the fourth bucket. The body's premise — that `implement-dispatch` refuses a `spec` issue outright — was falsified by #897 landing `--spec <n> --slots <k>` the same afternoon.

**AC3, measured live, both readers at one instant.** The reader as merged on `origin/main` prints 16 `unresolved`; this branch prints 15 `unresolved` and 1 `spec`. The two lines that differ are both #885, moving rather than vanishing:

```
-unresolved  885 Spec: implement-spec and burndown rebuilt on the herdr lane (map #776)  (no native dependencies and no `Blocked by` of any form)
+spec        885 Spec: implement-spec and burndown rebuilt on the herdr lane (map #776)  (a spec parent: dispatch with `implement-dispatch --spec 885 --slots <k>`)
```

The drop is exactly the number of spec issues in the queue. The ruling's "13 to 12" was a reading from earlier in the day; the board gained tickets between readings, and the claim it makes is unchanged.

**AC2 is answered by `needs-info` alone.** `implement-dispatch`'s refusal set (`flow/lane/src/bin/implement_dispatch.rs:465-489`) is `{in-progress, needs-info, spec}`. `in-progress` already left as a claim, `spec` is now its own bucket, so `needs-info` is the one label left for `NON_DISPATCHABLE_LABELS`.

**A spec parent whose stated blockers cannot be read stays `unresolved`.** Nobody ruled on this case. Silence is what was ruled into `spec`; a ticket that *states* prerequisites this reader cannot resolve has not been silent, and an unknown prerequisite is not a met one. The verification pass agreed and noted the file already commits to it — `classify`'s docstring says an unreadable state "is unresolved rather than a guess".

**`docs/agents/issue-tracker.md` is a file the ticket never named.** Its frontier query described three buckets and would have been wrong the moment this lands. Accepted by the controller.

**`burndown/SKILL.md` is deliberately absent** though it described three buckets too. `skills-898` held the file; the controller handed it the wording and it committed at `8be20cc`.

### Round-1 findings, every one with its disposition

- **S1** — fixed, `24b1044`. `classify` announced three buckets while returning four.
- **S2** — fixed, `24b1044`. Test module docstring, same staleness.
- **S3** — fixed, `24b1044`. **The find of the round.** `unresolved_count` reassigns `NON_DISPATCHABLE_LABELS` before each read, so the AC3 harness test never read the shipped value — it passed with `frozenset()` shipped. A test that cannot fail on the thing it names, sitting in the harness written to prove the fix. It now reads with no swap as well, and dies on its own under an empty set.
- **S4** — `disputed: a module-constant swap under try/finally is how a harness is parameterised in a sequential runner, which is what was asked for. The restore was verified by mutation, and it breaks only under a threaded in-process runner, which this suite is not.`
- **S-OE1** — `disputed: yagni on unresolved_count, same reason as S4.`
- **S-OE2** — `disputed: the ruling explicitly asked for the general rule stated in references/frontier.md, and the repo's house style endorses reason-giving comments in the file itself.`
- **P1** — fixed in this body: amended AC3's on-repo reading was recorded nowhere. It is under Decisions made above.
- **P2** — `disputed: AC4's literal "asserts it is dropped, not classified" is superseded by amended AC1, written in the same ruling comment.`
- **C1** — fixed, `a22834d`. The `spec` branch ran before the blocking sources, so a spec parent with an open blocker rendered in `spec` carrying a dispatch verb — the one reader whose job is never to offer work with an open prerequisite, handing over a nested run onto blocked work. No fixture covered it; every spec fixture in round 1 had a silent body. Four now do.
- **C2** — fixed, `24b1044`. Same docstring staleness as S1, plus `frontier`'s.
- **C3** — fixed, `24b1044`. No defect found; its one real note, a redundant trailing total-count assertion, is removed.

## Last reviewed sha

`a22834d`. One commit after it, `0c57bf7`, was not re-reviewed: a rename of a local flag (`declared` → `stated`) that the verification pass suggested as optional, plus the `Closes` trailer. No behaviour change.


caneff:

`codex-companion.mjs adversarial-review --wait --base origin/main`, ticket title + body + comments as context. Verdict: `needs-attention`, one finding. **Disputed on evidence Codex could not see.**

### DISPUTED — [medium] "Burndown controller instructions omit the new spec bucket" (`burndown/SKILL.md`)

The sentence exists. It is on `implement-898`, committed at `8be20cc`, and reads:

> `python3 burndown/frontier.py <owner/repo> <label>` prints the `unblocked`, `blocked`, `unresolved` and `spec` buckets. Native tracker dependencies first, the `## Blocked by` section as the fallback, and a ticket with neither is **unresolved** — never dispatched on the assumption that silence means clear. A `spec`-labelled parent is none of those three: it is dispatchable by a different verb, and its entry names that verb — `implement-dispatch --spec <n> --slots <k>` — so a controller can act on the line without opening another document.

Verified in that worktree at `d41a5fb`: the four-bucket list is present and `implement-dispatch --spec` appears in the file.

`burndown/SKILL.md` was held by another worker for this entire wave, so a second writer in it is the collision the lane exists to prevent. The wording is this PR's worker's own, drafted here and handed across at the controller's direction — the fourth such cross-ticket line in map #776 and the same trade made three times before it. Both PR bodies record the split.

Codex reviews one branch's diff against `origin/main` and cannot see another open branch, so the finding is correct about `main` today and wrong about the work. It is the second instance of exactly this false positive in this map — the same shape was disputed on PR #930, where the `merge-tail.md` pointer it reported missing was sitting in PR #929.

That recurrence is worth recording rather than just dismissing: **a review scoped to one branch cannot see a cross-ticket split, so any wave that splits a file across PRs will manufacture this finding once per split.** The controller verifies both halves are on `main` before agent-skills#899 closes the map, which is the check that actually catches it if a half is ever dropped.

### Controller notes

- The C1 precedence ruling is implemented as ruled and one step further than specified: the spec override fires only when `rank` returned `unblocked`, or `unresolved` with nothing stated, so `blocked` outranks `spec` **and** an unreadable blocker declaration stays `unresolved`. Silence is the case ruled into the bucket; a stated prerequisite that could not be resolved is not silence, and an unknown prerequisite is not a met one.
- The verification pass checked finding S3 in isolation — that test alone under an empty shipped set raises `AssertionError: frozenset()` — proving the witness is real rather than another test failing first and masking it.
- Body file built with `--json title,body,comments` and checked non-empty before the run.

No round 2: the one finding is disputed with evidence, and the ceiling is not spent on a false positive.

PR #943, opened 2026-09-20
multi-axis-code-review: capture the diff once, hand every axis the file (#937)

### LINKED ISSUE #937: multi-axis-code-review hands three axes a diff command, not the diff
`multi-axis-code-review/SKILL.md` line 37 captures "the diff **command** once" and then passes that command to each axis (lines 134, 140, 146). `flow/claude/agents/diff-reviewer.md` tells every axis to run it: "Run the diff command yourself — the caller's HEAD is not the branch under review, so use the `git -C <worktree>` form exactly as given."

So three opus agents each run the same `git diff <fixed-point>...HEAD`, then each cold-read the same changed files, and the standards axis additionally re-reads the same standards docs the others skip. The diff is identical for all three by construction — the fixed point is already pinned by the caller and `git -C <worktree>` makes it deterministic — so two of the three derivations are pure duplication before any reasoning starts.

What to change: the caller writes the diff to a file once and hands each axis that path alongside the command, so an axis reads rather than re-derives. Keep the command in the prompt as the provenance record and the fallback — an axis whose diff file is missing or empty should re-derive and say it did, rather than review nothing. The `git -C <worktree>` requirement exists because the caller's HEAD is not the branch under review; a captured file satisfies that requirement more directly than a re-run does.

Why it is worth doing: measured in `docs/research/2026-09-15-phase-timings.md` across 15 tickets, review round 1 is a median 3.7 minutes and round 1 plus verification is a median 40% of a ticket's whole dispatch-to-PR span. The three axes run in parallel, so the span is `max(axes)` and the duplicated reads sit on the critical path of every one of them.

What not to change while here: `model: opus` on the axes, and the correctness axis's witness check ("strip the constraint under test and see whether the assertion still passes"). The witness check is the most expensive part of a review and it is what catches hollow tests — three findings today (#893's `paths()` fail-open, #894's zero-cores default, #897's suppressed contradictions) came out of that discipline. This ticket is about duplicated I/O, not about reviewing less.

Filed from conversation — controller investigation of review latency during map #776.

caneff:
Closes #937

## What changed

The caller of `multi-axis-code-review` derives the diff once into
`<dir>/diff-<n>.patch`, beside the reports and sidecars, and every axis prompt
carries that path and its line count next to the command that produced it.
`flow/claude/agents/diff-reviewer.md` tells an axis to read that file rather
than re-run the command; the command stays as the provenance record and the
fallback, and an axis whose file is missing or empty re-derives and says so.
`multi-axis-code-review/diff-capture.test.sh` pins both homes, the way every
other `*-wording.test.sh` in this repo pins a skill's prose.

## Tests run

`bash tests/all.sh` — `63 suites passed`, exit 0.

## Decisions made

The seam under test is the prose of the two skill files. #937 named no seams,
and there is no harness that runs a skill's prose; 19 sibling
`<skill>/<topic>.test.sh` suites assert on it the same way, so I followed the
convention rather than escalating.

Round-1 findings, all eleven, each with its disposition:

- **S1** (`SKILL.md` § 4: angle-bracket placeholders put into a block that was
  runnable shell) — fixed, `4d61df2`.
- **S2** (the rationale restated in four homes) — fixed, `4d61df2`. Partly: § 1
  is one sentence now and the phase-timings citation is gone, but the
  verification pass counted the § 4 rationale growing 13 → 17 non-blank lines
  because the other fixes added normative content there. The rationale still
  lives in four places.
- **S3** (`[ -s ]` re-checking step 1's non-empty check) — fixed, `4d61df2`: the
  comment names the failed-or-empty *write*, and `|| exit 1` was added to the
  `git diff` line.
- **S4** (the test's `git -C <worktree>` needle matched the pre-change file
  verbatim, so it proved nothing) — fixed, `4d61df2`: it asserts the fallback
  sentence instead.
- **S5** (stale TDD comment `# no such prose there yet`) — fixed, `4d61df2`.
- **P1** and **C3** (the capture is keyed on `<n>` alone, so a second round
  reuses round 1's diff — present and non-empty, therefore invisible to the
  missing-or-empty fallback) — fixed, `4d61df2`: § 4 says to re-capture at the
  start of every round.
- **P2** (the axis bullets described "the captured diff file" instead of naming
  the path) — fixed, `4d61df2`.
- **P3** (the test asserts #937's own "what not to change" list) — disputed:
  that is a regression guard on precisely what the ticket warned off, and
  prose assertion is this repo's only seam over a skill's text.
- **C1** (`<n>` left unexpanded is a literal the write and the guard agree on,
  so the guard passes, the capture lands under a name no prompt points at, and
  every axis silently falls back — the round costs what it cost before) —
  fixed, `4d61df2`: shell variables, so an unexpanded placeholder is a parse
  error before anything runs.
- **C2** (`Read` stops at 2000 lines and a truncated patch looks like a whole
  one) — fixed, `4d61df2`: `wc -l` in the block, the count in every prompt, and
  the standing brief says to read to the end.

### Codex adversarial review, round 1 (controller pass)

- **CONFIRMED, high — the captured diff path was keyed on the issue number
  alone**, under a repo-wide shared directory, with a truncating write — fixed,
  `f2893dc`. A retry, a verification round begun while a lagging axis is still
  reading, or another worktree on the same issue replaced the file under the
  first invocation's readers, and the missing-or-empty fallback cannot see a
  replaced-but-non-empty patch: three axes review the wrong tree holding the
  previous round's commit list and spec context, with no error anywhere. The
  name now carries the short revision and a per-invocation nonce, the write
  goes to a `mktemp` file published with `mv`, and every axis of one invocation
  gets the exact path the block printed rather than a pattern. The 14-day sweep
  in the same block still collects them — more files, not more lifetime.

  Codex called it correctly as this ticket's own defect class one layer out:
  C1/S1 was a guard that tested for absence when the hazard was wrongness, and
  so was this.

  The regression test extracts the fenced block from § 4 and runs it, twice for
  issue 937 at two distinct revisions, asserting distinct paths, the first
  capture unchanged, the two byte-different, no temp file left behind, and a
  20-day-old capture under the new key swept. Reverted to the single shared
  key it fails on exactly the three assertions Codex predicted.

## Last reviewed sha

`4d61df2` — the verification pass ran against it. `f2893dc` came after it, as
the Codex round-2 fix, and was not re-reviewed by the three axes.


caneff:

`codex-companion.mjs adversarial-review --wait --base origin/main`. Verdict: `needs-attention`, one finding. **Confirmed from the branch source.**

### CONFIRMED — [high] The captured diff path is keyed on the issue number alone

```
dir="$HOME/.cache/agent-reviews/$(basename "$(dirname "$top")")"
n=<issue number from step 2, or the branch name>
git -C "$worktree" diff "$fixed_point"...HEAD >"$dir/diff-$n.patch" || exit 1
[ -s "$dir/diff-$n.patch" ] || exit 1
```

The cache key is the issue number under a repo-wide shared directory, and the write truncates. A second invocation for the same issue — a retry, a verification round begun while an axis from round 1 is still reading, or another worktree reviewing the same issue — replaces the file under the first invocation's readers.

The fallback does not catch it. `diff-reviewer.md` re-derives only when the file is **missing or empty**, so an axis handed a replaced-but-non-empty patch reads it as valid while still holding round 1's command, commit list and spec context. The result is a review that looks complete and is against the wrong tree, with no error anywhere.

That is this ticket's own defect class, one layer out. The round-1 finding C1/S1 on this branch was an unexpanded `<n>` placeholder that the write and the `[ -s ]` guard agreed on, so the guard passed and every axis silently fell back — the saving evaporating with no error. This is the same shape: a guard that tests for absence when the hazard is wrongness.

Fix: make the capture unique per invocation — an immutable revision or worktree identifier plus a round or nonce — and write to a temporary file, then publish atomically so a reader never sees a half-written patch. Pass that unique path to all axes of that invocation. Regression test: two captures for the same issue number from distinct revisions, proving the second does not disturb the first.

### Controller note — the sibling-branch context was supplied and no false positive followed

Per agent-skills#941, this pass's focus text named the one open sibling branch (`implement-898`), the files it holds, and stated that nothing in this PR is split with it. The two previous passes in map #776 each produced a cross-ticket false positive (PR #930's `merge-tail.md` pointer, which was in PR #929; PR #940's four-bucket sentence, which was on `implement-898`). This pass produced none. One trial is not a result, but it is the first evidence for agent-skills#941 and it cost one paragraph.

Issue #835, opened 2026-09-15
codex-fourth-axis-wording test fails after #827's .scratch/ note


caneff:
Part of #776. `bash implement/codex-fourth-axis-wording.test.sh` fails on `main`: `FAIL: implement/SKILL.md is missing (in the expected section): /tmp`. Post it as a PR comment before acting on it`. PR #827 (574da7c, ticket #822) inserted "; the `mkdir -p` above is required because the worker's own Before the PR step already deleted this directory." into § The merge step 3 between "never `/tmp`" and "Post it as a PR comment…", splitting the phrase the test pins (`check_in "$merge_section" '/tmp`. Post it as a PR comment before acting on it'`). Fix the wording or the test so both say what § The merge means, and run every `implement/*.test.sh` before PR up.

Filed from: controller verifying a worker's "pre-existing failure" note during #824, 2026-09-15.

caneff:
Add to scope (controller, 2026-09-15, from PR #833's final Codex run): `implement/SKILL.md` § Before the PR step 5 says each closing reference must be "in this repo — an entry's `repository` field pointing elsewhere doesn't count", then a few lines later tells the worker a cross-repo ticket needs `Closes owner/repo#<n>`. A cross-repo closer can never pass. `implement-dispatch` only claims tickets in the PR's own repo, so drop the cross-repo remedy (or define expected closers as owner/repo + number pairs). Codex output: https://github.com/caneff/agent-skills/pull/833

Issue #817, opened 2026-09-15
implement: the controller runs the Codex adversarial pass at PR up, not the worker


caneff:
Chris ruled (2026-09-15). The goal of #812 is an independent Codex review of a Claude-run pipeline. After #814 (PR #816) the pass is real (a fresh Codex thread that reads the branch diff itself), but it is not independent of the builder:

1. The worker it reviews runs it, reads it, and classifies its findings (`codex-only, confirmed` / `also found by Claude` / `disputed`), or reports "skipped". Nothing contradicts a wrong call.
2. The raw output is not kept: #814's run wrote it to `/tmp` and it is not in the PR.
3. Only prose enforces it.
4. Codex never sees the fixes: the verification pass is Claude-only.

## Wanted

- **Move the pass to the controller.** Remove the worker's Codex fourth axis from `implement/SKILL.md` § Heavy tier / § Review. The worker's round 1 goes back to the three `/multi-axis-code-review` axes. Remove the worker's trial-row step and its Decisions-made Codex classes. `codex-lane.md` (Codex builds) is out of scope: leave its review step as #816 left it.
- **Controller step in § The merge**, on a heavy Claude-lane PR, after the CLEAN check and before the merge. Use #816's invocation: `installPath` from `~/.claude/plugins/installed_plugins.json`, the ticket body written to a file and passed through one `"$(cat "$body_file")"`, and `--base origin/<default>` run from the PR's workspace. Write the output to the workspace's git-ignored `.scratch/`, never `/tmp`.
- **Record it.** Post Codex's raw output as a PR comment (`gh pr comment <pr> --repo <owner/name> --body-file <file>`) before acting on it.
- **Act on it.**
  - Verdict with no material findings → merge as normal.
  - Findings → hold the merge. Send the worker the findings with the comment URL. The worker disposes of each one (fixed in a commit / `disputed: <why>` / filed) and sends "PR up" again. The controller re-runs the pass once on the fixes. There is no third Codex run: the second run's findings get dispositions in the PR body, and then the controller merges.
- **Classify for the trial.** The controller, not the worker, assigns each Codex finding its class by comparing it with the PR body's round-1 findings, and appends the row to `docs/research/2026-09-14-codex-review-trial.md`. The row is an auto-ship commit on `main` after the merge (docs/research is not code). Update that doc's how-it-works text and the trial-end wording in `SKILL.md` so the controller counts rows and brings Chris the table after five.
[...]

Issue #911, opened 2026-09-20
/file-ticket emits a Blocked by section, so hand-filed tickets are readable


caneff:
## TL;DR

Tickets filed one at a time during a build never say what they are waiting on, so the tool that picks the next piece of work cannot classify them and a human has to sort them out by eye.

## What to build

`/file-ticket`'s issue template gains a `## Blocked by` section, emitted on every ticket it creates, carrying either the blocking ticket references or "None — can start immediately".

`/to-tickets` already does this — #890 made it emit native dependency edges where available and the `## Blocked by` section otherwise. `/file-ticket` is the other source of tickets in this lane and it was not covered, so it is now the only producer still emitting tickets the frontier reader cannot read.

Use the same grammar #890 documents, not a new one. If `/file-ticket` can determine the blocking edge at filing time — it usually can, since it is filing from a finding that arose inside a known piece of work — it should emit a native dependency edge too, the same as `/to-tickets`.

## Evidence

Measured by `skills-890` on this repo, 2026-09-20: `ready-for-agent` reads 6 unblocked, 7 blocked, 8 unresolved. Seven of the eight unresolved are parentless tickets filed ad hoc during builds rather than through `/to-tickets` — #869, #871, #874, #875, #880, #881, #900. Each has a routing label and no statement of what it waits on. (The eighth is the spec parent itself, #910.)

The same ratio appeared two repos apart: 8 of 21 `ready-for-agent` tickets on `sudokumaker-custom-constraints` during the #781 burn had no blocking statement at all, which is the measurement #890 was written from. Closing `/to-tickets` alone fixes the batch case and leaves the ad-hoc case, which is the one that grows fastest during a burn — every finding a worker files lands here.

Reading a missing section as "unblocked" is the expensive failure #890 exists to prevent; the reader now reports `unresolved` instead, which is correct but still costs a controller a decision per ticket. Closing the loop at this source is what makes the bucket shrink.

## Acceptance criteria

- [ ] `/file-ticket`'s created issue carries a `## Blocked by` section on every path
- [ ] The section uses the grammar #890 documents, with "None — can start immediately" when there are no blockers
- [ ] A native dependency edge is added when the blocker is known at filing time
- [ ] The skill's own instructions state it, so the section is not left to the filer's judgement
- [ ] A ticket filed by the skill classifies as `unblocked` or `blocked` by the frontier reader, never `unresolved`

## Blocked by

- #890 (it defines the grammar this emits and the reader that consumes it)

[...]

Issue #888, opened 2026-09-19
implement: make step 3's second Codex run conditional on new commits


caneff:
## TL;DR

After a reviewer raises a point, the lane always runs the reviewer a second time. But if the worker answered the point by arguing or by filing a follow-up ticket rather than by changing code, the second run reads a file that has not changed by a single byte — and returns the same answer, for several minutes and a token budget.

## What to build

`implement/SKILL.md` § The merge step 3 says that after the worker disposes of a finding, the controller re-runs step 2 "and then this pass once more on the fixes — there is no third Codex run, so whatever this second run finds is final". Written literally it mandates a second run whose input cannot have changed.

Make the second pass **conditional on new commits**. If every disposition is `disputed` or `filed` and the PR head sha is unmoved since the first pass, there is nothing to re-review: the controller records the dispositions in the PR body and proceeds to step 4. The step's own tail already permits controller-level disposition; this makes the entry condition match.

State it explicitly, so a controller following the letter does not spend a run confirming what it already knows.

## Evidence

Hit twice, both on #781's burn and its tail:

- **sudokumaker-custom-constraints#559** — the disposition was `filed`, the worker pushed nothing, and the diff stayed at `203ac7a`. The controller skipped the second run and disposed at controller level, deviating from the letter of the step.
- **agent-skills#877** — the single finding was one the worker had already filed as #880 *before the pass ran*. Same shape: nothing to send back, nothing to re-read, diff byte-identical at `b96aa32`.

## Why this blocks the burndown spec

A burn's controller runs this step once per landing. Ten landings in the trial meant ten passes over this branch; two of them hit the unchanged-input case. The spec (#885) makes the controller's merge tail a written loop rather than a per-PR judgement call, so a step that mandates a provably useless run gets executed rather than quietly skipped.

## Acceptance criteria

- [ ] The second pass runs only when the PR head sha moved since the first
- [ ] When it does not run, the controller records each disposition in the PR body and proceeds to step 4
- [ ] The "no third run" ceiling is unchanged
- [ ] The single `-- "$(cat "$body_file")"` substitution shape is unchanged
- [ ] `implement/codex-adversarial-invocation.test.sh` covers the unchanged-sha case and still passes

## Blocked by

None — can start immediately.

## Seams under test

- `implement/codex-adversarial-invocation.test.sh` — the existing guard on step 3's invocation shape; extend it with the conditional-second-run case.

PR #950, opened 2026-09-20
The Codex pass is told what the tree cannot say, and collected behind a fail-closed gate (#941, #942)

### LINKED ISSUE #941: The Codex pass is told nothing about sibling branches, so a cross-ticket split reads as a missing line
`codex-companion.mjs adversarial-review` reviews one branch's diff against `origin/<default>` and cannot see any other open branch. When a wave deliberately splits one file's changes across two PRs — because one worker holds the file and another needs a line in it — the pass reports the absent half as a missing requirement. It is right about `main` and wrong about the work, every time, once per split.

Twice on 2026-09-20 in map #776:
- PR #930 (agent-skills#896): "the canonical loop does not point to merge-tail" — the pointer was in PR #929, written there at the controller's direction because `burndown/SKILL.md` was held by the #894 + #895 clump.
- PR #940 (agent-skills#910): "Burndown controller instructions omit the new spec bucket" — the four-bucket sentence was on `implement-898` at `8be20cc`, verified present in that worktree.

Each cost a finding, a controller verification against the sibling worktree, and a written dispute on the PR. Neither was a defect.

What to change: the focus text the lane hands `adversarial-review` names the open sibling branches and what is split across them — which file, which line, which PR carries it — so the pass can tell a deliberate split from a dropped requirement. The focus argument is already free-form (`-- "$(cat "$body_file")"`), so this is a change to what the lane composes, not to the companion script.

Where: `implement/codex-lane.md`'s invocation block and `implement/SKILL.md` § Review, which build `body_file` from the ticket's title, body and comments. The split is known to the controller at dispatch time — it is the controller that orders a cross-ticket line — so it can be stated rather than inferred.

What this does not change: the controller still verifies both halves are on `main` before the closing ticket of a map lands. That check is what catches a half genuinely dropped, and it is not replaced by telling Codex about the split.

Filed from conversation — controller review of Codex pass cost during map #776.

### LINKED ISSUE #942: The Codex pass runs after PR-up, serialized, and always with --wait
Two scheduling costs in the controller's Codex adversarial pass, both independent of what the pass does.

**It runs after PR-up.** The controller launches `adversarial-review` once the worker reports its PR, so the pass is pure serial time bolted onto the merge gate — after three opus review axes and a verification pass have already read the same diff. If it launched when the worker reports round 1 is out, it would overlap the worker's own verification pass instead of following it.

The objection is that the diff can move under the pass, and the machinery for that already exists: agent-skills#888 built skip-or-rerun keyed on `headRefOid` plus a `sha256sum` of the rendered body-and-comments, precisely so a second pass is skipped when nothing moved and forced when something did. The same key answers "did the branch move while the pass ran".

**Every invocation uses `--wait`.** `codex-companion.mjs adversarial-review` accepts `[--wait|--background]` and the lane always takes `--wait`. With more than one PR queued — three were, on 2026-09-20 — the passes run one after another unless the controller remembers to background them by hand. They should all launch, then be collected through `status` / `result`.

What to decide and build: where in the lane the pass fires, and whether `--background` plus a collection step becomes the default shape for a controller holding more than one PR. State what happens when a pass's sha no longer matches at collection time — rerun, or report against the sha it read and say so.

Measurement gap worth closing in the same work: no per-pass duration has ever been recorded. The controller ran five passes on 2026-09-20 and can only bound them by output-file timestamps, because every run went to a background shell with no timing. Whatever lands here should record the pass duration the way `docs/research/2026-09-15-phase-timings.md` records the review phases, so the next person arguing about this has numbers.

Related: agent-skills#941 (the pass is told nothing about sibling branches), agent-skills#937, agent-skills#938, agent-skills#939 (duplicated and unscoped work inside the Claude-side review round).

Filed from conversation — controller review of Codex pass cost during map #776.

caneff:
Closes #941

Closes #942

## What changed

The focus text handed to `adversarial-review` now ends with a two-line controller-context appendix — the open sibling branches and what is split onto them, and the posture of the code when it is not the posture the tree implies — at both invocation sites, with the "nothing to report" case written out rather than omitted (#941).

The pass now launches at a new "Round 1 out" wake from the worker and is collected at the merge gate, which refuses to merge on a verdict that is absent, unreadable, errored, raced or stale rather than treating it as a pass (#942). Its files live in `~/.cache/agent-reviews/<repo>/`, not a `.scratch/` the worker deletes out from under an in-flight run.

Every run, collected or refused, records its own duration into `docs/research/2026-09-20-codex-pass-durations.md`; no per-pass duration had ever been recorded.

## Tests run

`bash tests/all.sh` — 68 suites passed at `9a58707`, rebased onto `437f59c`. Two of those suites are new here: `implement/codex-focus-appendix.test.sh` and `implement/codex-pass-schedule.test.sh`.

## Decisions made

Round-1 findings, all fourteen, with dispositions (sidecar: `~/.cache/agent-reviews/skills/dispositions-941.jsonl`):

- **S1** fixed in `a6e64ac` — `codex-lane.md` restated the appendix template under its own pointer and the copy had already drifted; it points now.
- **S2** disputed: § The merge's length is this file's deliberate shape — each rule carries the incident that earned it. The duplicated durations rationale was trimmed; the length stands.
- **S3** fixed in `a6e64ac` — the launch block now follows the render that builds `body_file`.
- **S4** fixed in `a6e64ac` — the subsumed second grep in the column check is gone.
- **S5** fixed in `a6e64ac` — the fail-open clause is stated once in SKILL.md.
- **P1** disputed: #941's body names § Review, but since #817 no adversarial invocation lives there, so an appendix written into § Review would attach to nothing. The ticket's pointer is stale, not its ask: the remedy went to § The merge step 3, the one place that builds `body_file`, which serves the ticket's intent rather than its wording — noted here so the discrepancy is legible to anyone reading #941 later.
- **P2** disputed, on three independent reads of the plugin source rather than of its docs — the spec axis, the correctness axis and the controller each checked `codex-companion.mjs` 1.0.6 separately: `handleReviewCommand` lists `background` in its `booleanOptions` and never reads it, calling `runForegroundCommand` unconditionally; `options.background` is branched on only in `handleTask`. The usage text advertises `[--wait|--background]` for `review` and `adversarial-review` all the same, so the flag is accepted, documented and inert — filed upstream as #951, not this lane's to fix. The lane says what the flag does today and backgrounds the shell instead; one pass in flight per PR is now stated, and each is a node process against the box cap.
- **P3** fixed in `a6e64ac` — the skip clause is bounded to the preflight; an errored run is a refusal that reruns.
- **P4** disputed: the durations file ships with no rows because the five passes of 2026-09-20 were never timed, and a bound taken from a file mtime is not a measurement.
- **C1** fixed in `a6e64ac` — a collected verdict is posted from the cache directory and left to its 14-day prune; the `.scratch/` cleanup is scoped to a pass run at the gate, so it can no longer delete the only copy of the verdict or `rmdir` the directory the Claude axes' reports live in.
- **C2** fixed in `a6e64ac` — the launch block defines `plugin_root` and `body_file`, which only existed in the block 110 lines later.
- **C3** fixed in `a6e64ac` — the record carries the node call's exit status, and a non-zero status is a refusal; a pass that failed and returned used to write a record passing every sha test.
- **C4** fixed in `a6e64ac` — the overlap is banked only on a round 1 whose findings produce no fix commit, now stated.
- **C5** fixed in `a6e64ac` — same fix as S4.

Codex round 2 raised one, fixed in `8e83d1d`: the gate-time execution path — the one that runs precisely when the early record is refused — was the legacy `.scratch/` block with no exit status, no record, no validation and no duration row, so a retry that errored would have been posted and read as "no material findings". There is now one block for every phase (`early`, `gate-retry`, `second`), the retry answers to the same five refusals, and a retry that is itself refused ends the step as a visible `Codex pass skipped: <why>` with no trial row. The pass's files leave the workspace's `.scratch/` for `~/.cache/agent-reviews/<repo>/`, which removes the hazard the `rm`/`rmdir` apparatus guarded (#835) rather than guarding it again; `implement/codex-fourth-axis-wording.test.sh` Rules 3 and 8-11 now pin the new location and carry the old incident as the reason it moved. The durations file gains `phase` and `collected-after-retry`, so a retried pass cannot be counted as a clean early one.

Verification pass raised two of its own, both fixed in `08ac5b9`: the C3 fix was only half-witnessed (stripping `status=$?` left both suites green), and the durations file's outcome vocabulary predated `errored` and never listed `absent`.

The overlap #942 buys is banked only when round 1 produces no fix commit; a fix pushed while the pass runs moves the head and the gate reruns it at merge. The durations file is what turns that from argument into arithmetic once a few passes have run.

Scope note: the diff touches `implement/SKILL.md`, `implement/codex-lane.md`, two new test suites and one new research doc — the files the tickets name, plus the durations file #942 asks for.

## Field evidence for the appendix

The Codex pass on this PR was the first use of § Controller context as written, verbatim, rather than phrased ad hoc. No sibling-branch or posture false positive followed — four passes running now with none. What that does *not* show is which shape is doing the work: the controller's three earlier hand-written appendices were imperative and this one declarative, and this pass produced neither kind of false positive, so nothing separates them yet. The rule that both lines are written even when empty is the part with an argument behind it, not a measurement.

## Also in this branch

§ Before the PR gains a step 4: read your own diff against the three recurring defect classes named in `AGENTS.md` § Recurring defect classes, with `docs/agents/defect-classes.md` for the checks and instances. A pointer, not a third copy; the two cross-references to the `gh pr view` check follow it to step 6.

## Last reviewed sha

`ad81b93` (the rebase onto `437f59c` rewrote every sha; this commit was `a6e64ac` when the axes read it). Three commits landed after it and were not re-reviewed by the Claude axes: `592e5fc` (the verification pass's own two findings), `63d0948` (Codex round 2's finding, which round 2 itself is the review of) and `9a58707` (the defect-classes pointer).


caneff:
## Controller: Codex adversarial pass — round 1, one finding, CONFIRMED

`codex-companion.mjs adversarial-review --wait --base origin/main`, focus text built with `--json title,body,comments` plus this PR's own § Controller context template, used verbatim as its first field test. Verdict: `needs-attention`, one finding. **Confirmed from the branch source.**

### CONFIRMED — [high] The refused-early retry runs through the ungated legacy path (`implement/SKILL.md` § The merge step 3)

The early-launch block introduced by this PR captures `status=$?`, writes `record` with ticket, status, `launch_sha`, `completion_sha`, `body_sha256`, `started` and `completed`, and feeds the duration row. The gate's foreground block, which is what runs when an early record is **absent, unreadable, errored, raced or stale**, is the legacy one: it writes `out_file` under `.scratch/` and nothing else. No exit status captured, no record written, no sha or body-hash validation, no duration action.

So the path that exists precisely to handle a degraded early pass is the one path with none of the checks. A retry that errors produces an `out_file` that gets posted and then read under the old "No material findings → go to step 4" flow, and a non-zero exit is invisible because nothing looked. The fail-closed collection this PR builds is enforced on the healthy path and skipped on the sick one.

The prose above it already states the stronger rule — "**Every run records its duration**, collected or discarded" — so the file argues against itself: a requirement stated for every run, with a mechanism attached to one of the two blocks that can perform it.

This is the **fifth instance in this trial of a stated fallback with no mechanism behind it**, after #888's incidentally-justified rule, #877's guard that cannot tell a citation from a use, #911's `--blocked-by` fallback no code path reached, and #897's sha-list review procedure no skill could execute. A rule is not shipped until something can carry it out.

Fix as Codex recommends: one recorded-and-validated execution path parameterized for where it launches, rather than two blocks with different guarantees. Require a valid record before posting or proceeding. Append the duration row for the rerun. Test a refused early record followed by a foreground retry, including a non-zero exit and the missing- and stale-record cases — that is the combination neither block's current tests construct.

### Controller note — the template's first field test

This pass carried § Controller context exactly as this PR defines it, including its rule that both lines are written even when empty. That rule is better than what I wrote by hand across the previous three passes: I phrased each appendix ad hoc, and an omitted line and a "nothing is split" line are indistinguishable to Codex. No sibling or posture false positive followed, which is now four passes without one.

I cannot tell you whether the declarative shape matters versus the imperative "do not report a finding whose remedy lives on another branch" I had been adding — this pass produced neither kind of false positive, so nothing separates them. Not evidence either way.

Issue #949, opened 2026-09-20
merge-cleanup: give an ignored status entry a type instead of a trailing-slash flag


caneff:
`WorktreeFiles` in `flow/lane/src/bin/merge_cleanup.rs` carries each `git status --porcelain` entry as a bare `String`, and the trailing `/` on that string *is* the "this is a directory" flag. Nothing names the convention, so every reader reconstructs it: `classify_ignored` tests it (`entry.ends_with('/')`), then the printed line and the walk both strip it, and `grep -c "trim_end_matches('/')" flow/lane/src/bin/merge_cleanup.rs` now returns six sites.

That the convention is implicit is not theoretical here. The correctness axis on PR for #869/#946 found that git quotes a porcelain path holding a non-ASCII byte, a quote, a backslash or a control character (`core.quotePath`), and the quoted form ends with `"`, not `/` — so a quoted directory read as a plain file and was never walked. That was fixed at the parse boundary by `unquote`, but the class of bug is the one an unnamed primitive invites: two encodings of "directory" that only a careful reader tells apart.

Suggested direction: a small type parsed once in `WorktreeFiles::read` that holds the decoded path and whether git reported it as a directory, so no downstream site re-derives either from string shape. `is_cache`, `classify_ignored`, `dirty_text`, `first_names` and the empty-directory line are the consumers.

Filed from the standards axis of the multi-axis review on the #869 + #946 build (finding S1... S3), raised as Primitive Obsession and disputed there as too wide for that diff.

## Blocked by

- #946

Issue #838, opened 2026-09-15
merge-cleanup: a non-cache ignored name like .scratch/ can still be elided behind modified/untracked files


caneff:
Part of #776. Blocked by #823 (same file, merge_cleanup.rs).

Found by the Spec-axis reviewer on #823's PR. `dirty_text` (`flow/lane/src/bin/merge_cleanup.rs`) chains its name list as `modified, untracked, ignored` before calling `first_names` (`NAMES_SHOWN = 5`):

```
let names: Vec<String> = self.modified.iter().chain(&self.untracked).chain(&self.ignored).cloned().collect();
format!("{} file(s) would be lost: {}", kinds.join(", "), first_names(&names))
```

Ignored entries (the scarcest, least-recoverable class — `.scratch/` evidence) sit last. A worktree with 6+ modified/untracked files plus `.scratch/` prints `.scratch/` nowhere — it falls into "and N more", the exact failure #823's ticket described (there for the cache-vs-ignored pair; here for modified/untracked vs ignored). #823 fixed the cache-label half; this is the same elision on the tracked/untracked side.

Fix direction: chain `ignored` first (or otherwise guarantee non-cache ignored names always show), consistent with #823's "what Chris approves losing is never in the elided tail." Ruling on whether modified/untracked also deserve priority over ignored is for whoever builds this — flag it, don't assume.

Where: `flow/lane/src/bin/merge_cleanup.rs`, `WorktreeFiles::dirty_text`.

Issue #823, opened 2026-09-15
merge-cleanup: distinct labels for cache vs ignored counts, non-cache names first


caneff:
Part of #776. Blocked by #821 (same file). `merge-cleanup implement-368 --discard` said '1 ignored file(s) would be lost: .scratch/' then 'discarding 15 ignored file(s)', with `.scratch/` hidden in 'and 10 more'. Two different counts share the label "ignored file(s)": the non-cache refusal list and the caches print (`{} ignored file(s) in {wt}` over `files.caches`). Call caches "cache file(s)", and have `first_names` (`NAMES_SHOWN`) list non-cache ignored entries like `.scratch/` before caches and `.venv/`, so what Chris approves losing is never in the elided tail.

Where: `flow/lane/src/bin/merge_cleanup.rs`, `first_names` and the caches `safe_println!`.
Evidence: #781 hurt 18.

Filed from: conversation triaging the #781 hurt list, 2026-09-15.

caneff:
Add to scope (controller, 2026-09-15): `merge-cleanup implement-822` refused with "1 ignored file(s) would be lost: .scratch/" when `.scratch/` was an **empty directory** — left by § The merge step 3's Codex pass, which creates `.scratch/` for its two files and removes only the files. An empty ignored directory holds nothing to lose; it should not refuse (or step 3 should `rmdir` the directory when it created it). Needed a manual `rmdir` before cleanup would run.

caneff:
Add to scope (controller, 2026-09-15, from PR #829's final Codex run): when `merge-cleanup`'s claim-clearing `gh issue edit` fails after the git cleanup succeeded, it prints the re-run command on stderr and exits 0, so a caller sees success while the closed issue keeps `in-progress`. Exit non-zero with a message saying git cleanup completed and claim clearing needs the printed re-run; contract test that the failing-edit path is non-success. Codex output: https://github.com/caneff/agent-skills/pull/829

caneff:
[...]

Issue #801, opened 2026-09-14
merge-cleanup: refuse ignored non-cache files like .scratch/ unless --discard


caneff:
`merge-cleanup <branch>` removes a worktree's ignored files without refusing: `guard_files` in `flow/lane/src/bin/merge_cleanup.rs` ("Ignored files never refuse; their count and first names are printed, since they go too"). On 2026-09-14 the controller's cleanup of `implement-774` discarded the worker's `.scratch/` that way — gone, not recoverable. The same file already disagrees with itself: `stale_worktrees` holds a worktree back because "ignored files included — .scratch evidence is work too". RULES.md sends long-run scripts, logs and outputs to a git-ignored scratch dir inside the worktree, so `.scratch/` is exactly where evidence lives.

Ruled (Chris, 2026-09-14, "fix the cleanup skill for that"): split ignored entries in two.
- **Regenerable caches** — a fixed allowlist of names: `node_modules/`, `__pycache__/`, `target/`, `.venv/`, `.pytest_cache/`, `.ruff_cache/`, `.mypy_cache/`. Removed with the worktree, printed as today.
- **Every other ignored entry** (`.scratch/` included) is treated like modified/untracked files: the single-branch form refuses, naming them, and `--discard` removes them anyway. `--sweep` lists such a worktree "dirty, not removed" even with `--yes`, as it does for dirty files now.

Update `--help` and the `guard_files` doc comment to say so. The allowlist is deny-by-default on purpose: an unknown ignored name is kept, because a wrongly kept cache costs a `--discard`, and a wrongly discarded note cannot be undone.

## Seams under test
- `flow/lane/tests/merge_cleanup.rs`: a worktree holding only cache entries (`node_modules/`, `__pycache__/`, `target/`) is removed; one holding `.scratch/` refuses, naming it; the same with `--discard` removes it; `--dry-run` reports the refusal without touching anything; `--sweep --yes` does not remove a worktree with `.scratch/`.

Filed from conversation, 2026-09-14.

PR #952, opened 2026-09-20
merge-cleanup: judge an ignored directory by what it holds, not by its entry

### LINKED ISSUE #869: merge-cleanup refuses on an empty ignored directory
## What

`merge-cleanup` refuses to remove a linked worktree whose only "at risk" content is an **empty** ignored directory, and demands `--discard`. Nothing would be lost. Because `--discard` is denied to agents by the auto-mode classifier ("Irreversible Local Destruction"), every such merge ends with a paste line for Chris for no reason.

## Evidence

In `caneff/twitch-rules-scroller`, every implement worktree gets an empty `e2e-artifacts/` (ignored via `.gitignore`). On 2026-09-16 that blocked the plain cleanup of **six** merged worktrees in a row — implement-274, -296, -275, -306, -301, -307 — each needing Chris to run `--discard` by hand:

```
merge-cleanup: refusing to remove …/.claude/worktrees/implement-301 — 1 ignored file(s) would be lost: e2e-artifacts/ (--discard overrides)
```

The directory held nothing, checked each time:

```
$ git -C …/implement-301 status --porcelain --untracked-files=normal --ignored=matching
!! e2e-artifacts/
$ find …/implement-301/e2e-artifacts
…/implement-301/e2e-artifacts
```

`find -type f` returns 0. With `--ignored=matching`, git reports an ignored directory as one `!!` entry even when it contains no files, and the parse in `flow/lane/src/bin/merge_cleanup.rs` (the `"!!" => files.ignored.push(name)` arm, and `has_work` / the "would be lost" message built from it) counts it as work.

## Fix

When an ignored entry is a directory, count it only if it contains at least one file anywhere beneath it. An ignored directory holding no files, or only empty subdirectories, is not work and must not block removal.

Keep everything else about the deny-by-default ignored rule (#801) exactly as it is: a non-empty ignored directory with an unknown name is still kept, and `--discard` still overrides only for real content.

## Acceptance criteria

- [ ] A worktree whose only non-cache ignored entry is an empty directory is removed by plain `merge-cleanup`, without `--discard`
- [ ] The same, with an ignored directory that holds only empty subdirectories
- [ ] An ignored directory holding one file still refuses and names it, unchanged — so the #801 safety is witnessed as intact
- [ ] Tests in `flow/lane/tests/merge_cleanup.rs` cover all three
- [ ] The doc comment at the top of `merge_cleanup.rs` ("A linked worktree with modified, untracked or ignored files is never removed") says empty ignored directories don't count

## Why not a permission rule instead

Allowing agents `--discard` would remove the safety for the case it exists for — a worktree holding real unsaved evidence. The refusal here is a false positive, and the fix belongs in the check.

Filed from twitch-rules-scroller controller session, 2026-09-16, with Chris's go-ahead.

### LINKED ISSUE #946: merge-cleanup: an ignored directory is judged by its name, not its contents
`merge-cleanup`'s dirty-file guard treats one collapsed ignored directory as "1 ignored file", whatever is actually inside it. That is wrong in both directions: an **empty** ignored directory forces a needless `--discard` through Chris's hands, and a **full** one reports "1 file(s) would be lost" when it holds hundreds, so `--discard` is approved against a count that understates the loss.

Where: `WorktreeFiles::read` in `flow/lane/src/bin/merge_cleanup.rs` runs `git status --porcelain --untracked-files=normal --ignored=matching` and pushes each `!!` line onto `ignored` as one entry. `is_dirty` and `dirty_text` then count entries, not files.

Evidence: `--ignored=matching` does **not** expand a directory matched by its own `.gitignore` line. Repro in a scratch repo whose `.gitignore` holds `e2e-artifacts/`:

```
$ mkdir -p e2e-artifacts/sub && git status --porcelain --ignored=matching
!! e2e-artifacts/
$ touch e2e-artifacts/shot.png && git status --porcelain --ignored=matching
!! e2e-artifacts/
```

Identical output, zero files versus one. Observed live on 2026-09-20: cleanup of `implement-332` in `twitch-rules-scroller` refused with `1 ignored file(s) would be lost: e2e-artifacts/` while `find … -type f` returned 0. `npm test` there creates the directory on every run, so this fires on most cleanups in that repo.

Build: for each `!!` entry naming a directory, walk it on disk and classify by what it holds, not by the entry.

- Zero files (empty, or only empty subdirectories) → a new `empty` category. Not dirty, so cleanup proceeds with no `--discard`, and the worktree removal takes the directory with it.
- One or more files → ignored, with the true file count and the real names in `dirty_text`, so the refusal states the actual loss.

Chris's ruling on reporting: the empty directory is removed, and the run prints one line naming it — same shape as the existing `removing the linked worktree at <path>` line — so the output stays a full account of what cleanup touched. Do not silently drop it.

Keep the deny-by-default posture of #801: only a genuinely file-free directory becomes `empty`. An unreadable directory stays ignored and still refuses. `CACHE_DIRS` / `is_cache` classification is unchanged and still runs first.

Tests belong beside the existing guard tests: an empty ignored directory does not refuse and is named in the output, a full one refuses with its real count, and a directory holding only empty subdirectories counts as empty.

Filed from conversation (controller session, twitch-rules-scroller #332 cleanup).

## Blocked by

caneff:
Closes #869

Closes #946

## What changed

`git status --ignored=matching` collapses an ignored directory to one `!!` entry whether it holds nothing or hundreds of files, and the guard counted entries — wrong in both directions. `classify_ignored` now walks each non-cache `!!` directory on disk: no file anywhere beneath it becomes a new `empty_dirs` category, not dirty, removed with the worktree and named on its own line once the removal has actually happened; one or more lands in `ignored` under real names, so the refusal states the true count and the true loss.

A new `unquote` decodes git's `core.quotePath` C-quoting at the parse boundary, so the walk, `is_cache` and every printed name see the path git means rather than its transport encoding.

#801's deny-by-default posture is unchanged: `is_cache` still classifies first, and a directory the walk cannot read is not known to be empty — every `read_dir`/`file_type` error propagates as `None`, the entry stays `ignored`, and it still refuses.

## A test was deleted

`an_empty_ignored_directory_still_refuses_without_discard` is gone. It encoded #823's ruling that emptiness earns no exemption, which #869 and #946 reverse — both filed with Chris's go-ahead, #946 stating the reporting ruling in his words. Deleting it is the mechanical consequence of building these tickets; keeping it would ship a suite that fails the feature. `guard_files`' doc comment now records what #823 argued and why it was narrowed rather than abandoned, so the reversal does not read as a bare preference later.

## Tests run

`cargo test --manifest-path flow/lane/Cargo.toml` — 110 passed, 0 failed in `--test merge_cleanup`; whole `lane` crate green.

Every new test was mutation-checked. The load-bearing one for #801: `None => self.empty_dirs.push(entry)` — letting an unreadable directory read as empty — turns `an_unreadable_ignored_directory_still_refuses` red. That mutation is the witness that a permission error cannot approve a destructive removal. The others: a non-descending walk reddens the empty-subdirectories and full-directory tests; keeping the collapsed entry reddens the full-directory test; `unquote` as the identity reddens the quoted-directory test; printing the line back inside `guard_files` reddens the phantom-removal test; decoding a rename line whole reddens the quoted-rename test.

## Decisions made

Round-1 findings, all ten, each with its disposition. The verification pass confirmed every one and wrote `dispositions-946.jsonl`.

- **S1** — fixed in `17d757b`.
- **S2** — `disputed`: the duplication is real, but the other three sites are pre-existing and unifying a wording convention would widen a correctness diff.
- **S3** — filed as **#949** (`ready-for-agent`, blocked-by #946).
- **S4** — fixed in `17d757b`.
- **P1** — fixed in `17d757b` (same fix as C2).
- **P2** — `disputed`: the expectation change is the intended consequence of #946's real-names requirement, and the spec axis itself ruled intent preserved.
- **P3** — fixed in `17d757b`.
- **C1** — fixed in `17d757b`. The strongest finding of the round: git quotes any path with a non-ASCII byte and the quoted form ends with `"`, not `/`, so a quoted directory was filed as a plain file and never walked. Both bugs survived intact for such a name, including the direction #946 calls dangerous.
- **C2** — fixed in `17d757b`.
- **C3** — `disputed`: expansion changes no refusal decision, only message length. The never-cap posture is settled (#838/#847) and the stderr-line-shape concern is a separate follow-up the controller filed.

Also decided: `unquote` decodes at the parse boundary rather than only inside `classify_ignored`, so all four categories report real names instead of one category disagreeing with the others.

## Last reviewed sha

`17d757b` — the verification pass's fixed point. `61465e5` came after it and was **not** re-reviewed: it fixes V1, the defect that pass itself found.


caneff:
## Controller: Codex adversarial pass — round 1, one finding, DISPUTED

`codex-companion.mjs adversarial-review --wait --base origin/main`, focus text built with `--json title,body,comments` plus § Controller context (sibling branches and posture). Verdict: `needs-attention`, one `[high]` finding: **"Empty-directory approval is stale by the time forced removal runs."** A process can create an ignored file between `files_under`'s walk and `git worktree remove --force`, and the forced removal then deletes it without `--discard`, while the output reports only an empty directory.

**Disputed. The race is real, it is already filed as #879, and this PR narrows it rather than opening it.**

### It is not new, and it is not this diff's

#879 — *"merge-cleanup: close the window between the workspace guards and git's force removal"* — is open and describes exactly this: `cleanup_branch` runs `guard_files`, then `guard_live`, and only then `git worktree remove --force`; anything appearing between the last guard and that command is discarded rather than refused, because `--force` removes git's own final dirty-worktree check. Its body states the blast radius as every caller of `merge-cleanup` including `--sweep`, and names the two candidate fixes. It is labelled `needs-info` precisely because choosing between them is a grilling, not a build.

#879 was itself raised by **the Codex adversarial pass on PR #878**, disputed there on the same grounds, and filed rather than built. So this is the same finding, from the same reviewer, on its second PR — a class of false positive the appendix does not yet cover: **the pass cannot see the tracker, so an already-filed, already-disputed defect is re-raised as a no-ship on the next diff that touches the same file.**

### The direction is wrong

Before this PR, an empty ignored directory **refused** and required `--discard` from Chris's own hands. The forced removal still followed, so the check-then-act window was identical — with a human approval step inserted in the middle of it, which makes the window *longer*, not shorter: seconds of process time become minutes of human time, and #869 records six such refusals in a row on 2026-09-16.

After this PR, the same directory is removed without that pause. The window shrinks to the ordinary guard-to-removal gap that every other category has always had, and that #879 owns for all of them.

### What this diff does do about the class

`files_under` is fail-closed by construction and its docstring states the rule: *"`None` when any part of the walk cannot be read — the caller must not read an unreadable directory as an empty one."* Every `read_dir`, `DirEntry` and `file_type` error propagates as `None`, the entry stays `ignored`, and the removal still refuses. The witness is the `None => self.empty_dirs.push(entry)` mutation turning `an_unreadable_ignored_directory_still_refuses` red. That is the half of the hazard this ticket can close without touching the lane-wide removal protocol, and it is closed.

Codex's recommendation — "retain the refusal/`--discard` requirement for such directories" — is precisely the behaviour #946 was filed to remove, with Chris's explicit ruling in the ticket body. An explicit ruling outranks a reviewer's preference.

### Recorded

Filed nothing new; #879 already holds it, and I have added a note there that it has now been raised twice by the pass. No change requested on this PR. Merging.

Issue #944, opened 2026-09-20
The § cross-reference gate reports real misses and exits 0


caneff:
`tests/section-references.test.sh` prints unresolved `§` references against real repo files and then reports `ok` with exit 0, so `tests/all.sh` counts it as a passing suite. A gate that names a defect and passes anyway is not a gate.

Reproduced on `main` at `52288827`:

```
$ bash tests/section-references.test.sh
source.md:4: bare § Build follows target.md
source.md:3: § Step 3 for the plan not found in target.md
implement/codex-lane.md:27: § Build not found in implement/SKILL.md
implement/codex-lane.md:42: § Build not found in implement/SKILL.md
ok
$ echo $?
0
```

The first two lines are the suite's own fixtures (`source.md`, `target.md`) and are expected. The last two are real: `implement/codex-lane.md` cites `§ Build` in `implement/SKILL.md` twice, at lines 27 and 42, and that section does not exist there. Both the stale references and the exit code are defects, and the exit code is the worse one — it is why the stale references survived.

Two things to fix, and they are separable:

1. The suite fails when a real-file reference does not resolve. Its fixture cases must stay passing, so the exit code has to distinguish fixture output from repo findings rather than counting lines.
2. The two `§ Build` citations in `implement/codex-lane.md` resolve — either to the section that replaced `§ Build` in `implement/SKILL.md`, or by correcting the reference. Read the file before renaming anything; the convention in the tree governs.

Related and not the same: agent-skills#913 (the `§` step locator leaves three gaps the gate cannot see) and agent-skills#915 (the syntax the gate enforces is undocumented). Those are about what the gate can detect. This one is about the gate not failing on what it already detects — a hollow witness at suite level, the same shape agent-skills#910's round 1 found inside a single test.

Found by `skills-937` during map #776 and reported rather than fixed, correctly — it is pre-existing on `main` and outside that ticket's diff.

Filed from conversation — controller verification of that report.

caneff:
## Closing as not-a-defect: the premise was a controller misreading

`skills-944` disputed this ticket's premise before building, and it is right. I verified all three of its claims myself.

[...]

PR #956, opened 2026-09-20
tests/all.sh fails a suite that reports findings and exits 0 (#954)

### LINKED ISSUE #954: The gate trusts exit status alone, so a suite that reports failures and passes is counted green
`tests/all.sh` judges every suite by exit status and nothing else:

```
if out=$($cmd 2>&1 </dev/null); then
  echo "PASS $label"
```

`out` is captured and then discarded on success. So a suite that prints real failures and exits 0 is counted green and its output is never shown. That is defect class 1 in `docs/agents/defect-classes.md` — an absent or malformed answer read as a benign one — sitting in the harness whose entire job is to answer "is this tree good".

Live instance, reproduced on `main`, filed separately as #944:

```
$ bash tests/section-references.test.sh
implement/codex-lane.md:27: § Build not found in implement/SKILL.md
implement/codex-lane.md:42: § Build not found in implement/SKILL.md
ok
$ echo $?
0
```

Two real misses, the word `ok`, exit 0, `PASS` in the gate.

**Build:** the gate fails a suite that exits 0 while its captured output carries a failure signature, and names both the suite and the matching line. The signature set is the decision to get right — `^FAIL`, `^Traceback`, `not found`, `does not exist`, `missing`, `ERROR` are the candidates, and some suites legitimately print those words while narrating a mutation check that is expected to go red. Anchoring to line starts, or having a suite opt out by declaring its own vocabulary, are both acceptable answers; picking one and stating why is part of the ticket.

Fix #944 in the same change so the new check has a real instance to catch: `tests/section-references.test.sh` must exit non-zero when it reports a miss, and its two stale citations in `implement/codex-lane.md` are corrected.

**The test that matters:** a fixture suite that prints a failure line and exits 0 must fail the gate, and a fixture suite that prints the same words as part of an expected-red narration must not. Both, or the check is either useless or unusable. Mutation-check both directions.

Why this ticket rather than more prose: `AGENTS.md` § Recurring defect classes now asks every agent to check for this class by hand. A rule with no mechanism behind it is defect class 2, so the section is currently an instance of the thing it warns about. This is the mechanism for class 1 in the one place every suite passes through.

Filed from conversation — controller, after the class list landed at f1d6258.

caneff:
Closes #954

## What changed

`tests/all.sh` judged every suite by exit status alone — it captured each suite's output into `out` and discarded it on success — so a suite that printed real failures and exited 0 was counted `PASS` and its output was never shown. It now also fails a suite that exits 0 while its output opens a line with `FAIL`, `Traceback`, or `ERROR`, naming the suite and the line that matched.

`tests/output-scan.test.sh` and four fixtures under `tests/fixtures/output-scan/` witness both directions: three suites that print one signature word each at column 0 and exit 0 (must fail the gate), and one that prints all three mid-line while narrating an expected-red mutation check (must pass).

**There is no suite in this repo currently exhibiting this defect.** The guard is built against fixtures deliberately and predates its first incident. `AGENTS.md` § Recurring defect classes asks every agent to check for defect class 1 by hand; a rule with no mechanism behind it is class 2 in the same doc, so that section was an instance of the thing it warns about. This is the mechanism, in the one place every suite passes through.

## Tests run

`bash tests/all.sh` — `67 suites passed`, exit 0, run on a clean tree at `118b28b`.

## Decisions made

**The signature set, anchored at column 0.** `^(FAIL|Traceback|ERROR)`, with no per-suite opt-out mechanism. The ticket left the choice open between anchoring and a declared-vocabulary opt-out; a live counter-example decided it. `tests/section-references.test.sh` prints `not found` four times as legitimate narration while its own mutation check goes red by design, so a substring set containing `not found` would have fired on it the day it was written. Running every suite and grepping its output for the loose set (`not found`, `missing`, `does not exist`) matched 14 suites that were green; the anchored set matched none of them.

**No opt-out mechanism.** The anchor is its own escape hatch: a suite that must reprint a child's `FAIL` line while passing indents it, which is also how a human reader tells a quoted failure from the suite's own verdict. `tests/all.test.sh` is the only suite reprinting `^FAIL` today, and only on its failure path, so nothing needs the hatch yet. A declaration mechanism shipping with no user is a bypass that gets routed around, and a typo'd declaration silently reading as "scanning off" would be defect class 1 inside the guard against class 1. If a legitimate need appears, the ticket to add it will have a real instance behind it.

**One positive fixture per signature word, not one printing all three.** The gate greps with `-m1`, so a combined fixture keeps matching on the first word and deleting either of the others from the set goes unnoticed.

**The logging false positive is answered at the point of failure, not in a comment.** Python's `logging.error()` writes `ERROR:root:<msg>` at column 0, and `2>&1` folds stderr into what is scanned, so an ordinary `*_test.py` exercising a logging path is failed by this scan and cannot reach the indent hatch. Testing an error path is ordinary work, so this is a real false positive rather than a hypothetical one. It does not reopen the no-opt-out decision: both ways out already exist — assert on the log in the test (`assertLogs`, `caplog`), or give the logger a format that does not start at column 0 — and the better one is what a well-written test does anyway. The gap was discoverability, not capability, so `logging_remedy` prints both remedies when the gate rejects a line, and only when that line has the shape of a logging record. Unconditionally would point away from a real failure on a genuine `FAIL:` catch.

**The ticket's `fix #944 in the same change` instruction is void.** #944 was closed as not-a-defect during this build: its four "unresolved references" are `tests/section-references.test.sh`'s own expected-red narration, `implement/SKILL.md:129` still holds `### Build` (added at `f9eed03`, never removed) so the `implement/codex-lane.md` citations always resolved, and that suite cannot exit 0 on a genuine miss because it runs the checker under `set -euo pipefail`. Nothing in this diff touches either file.

### Round-1 findings

Every finding from the single round of `/multi-axis-code-review`, by the id its sidecar assigned. Nine fixed in code; none filed as a follow-up ticket. `P1` asked for a statement in the PR body, so it was unmeetable until this body existed — the verification pass recorded it in `dispositions-954.jsonl` under the `disputed` shape, because the schema has no "outstanding" one and `fixed` with a sha would have been a false claim.

| id | axis | disposition |
|---|---|---|
| `S1` | standards (hard) | fixed in `1a4ecc5` |
| `S2` | standards | fixed in `1a4ecc5` |
| `S3` | standards | fixed in `1a4ecc5` |
| `S4` | standards | fixed in `1a4ecc5` |
| `P1` | spec | satisfied by this PR body — the "no live instance" statement above |
| `P2` | spec | fixed in `1a4ecc5` |
| `C1` | correctness | fixed in `1a4ecc5` |
| `C2` | correctness | fixed in `1a4ecc5` |
| `C3` | correctness | fixed in `1a4ecc5` |
| `C4` | correctness | fixed in `1a4ecc5` |

`S1` and `C1` are the same hole found independently by two axes: only `FAIL` had a positive fixture and only `ERROR` appeared in the narration fixture, so the set could be narrowed to `^(FAIL)` with the whole suite still green — and the fixture's own comment claimed coverage it did not have. Two thirds of the decision the ticket called "the decision to get right" was unwitnessed. `S2`, `C3` and `C4` were unchecked `mkdir`/`cp` calls in the test harness, where a missing fixture would have produced a gate run over no suites, an exit of 0, and a red result naming the wrong cause.

One residual on `S3`, left in place deliberately: the remedy it named was applied (the test header points at the rationale block instead of restating it), but the "one fixture per word, because the gate greps `-m1`" paragraph now appears twice near-verbatim, in `tests/fixtures/output-scan/silent-traceback.sh` and in the `tests/output-scan.test.sh` header. Duplication was reduced where the finding pointed and partly moved rather than eliminated. The verification pass judged it not worth another round, and an unreviewed commit to fix a cosmetic residual is a worse trade than leaving it on the record here.

After the fix, every mutation goes red and names the fixture that caught it:

```
narrow to ^(FAIL)      -> RED, silent-traceback passed
narrow to ^(Traceback) -> RED, silent-failure passed
narrow to ^(ERROR)     -> RED, silent-failure passed
drop the ^ anchor      -> RED, narration suite rejected
disable the scan       -> RED, silent-failure passed
```

### Codex adversarial review, round 2

One finding, confirmed: the logging false positive above. Disposition — **fixed in `118b28b`**, and the reasoning is in the Decisions section: a message rather than a mechanism, because both escapes already existed and only discoverability was missing.

The test asserts the remedy in both directions. Checking only that the note appears for a logging-shaped line would pass just as well if the gate printed it on every failure, which is the hollow-witness shape this whole change exists to stop; so the `FAIL:` and `Traceback` cases assert the note is absent, and the marker phrase is text no fixture prints. Mutations, each red for its own reason:

```
remove the remedy note        -> RED, "without naming the way out (silent-error)"
make the remedy unconditional -> RED, "not a logging record (silent-failure)"
drop the call site            -> RED, "without naming the way out (silent-error)"
```

## Last reviewed sha

`118b28b`. Round 1 reviewed `92c459d`; the verification pass covered the round-1 findings and the fix commit `1a4ecc5`; the Codex round-2 pass read `1a4ecc5` and its one finding is fixed in `118b28b`. There is no third pass, so `118b28b` itself carries only the round-2 fix and its mutation-witnessed test.


caneff:
## Controller: Codex adversarial pass — round 1, one finding, CONFIRMED

`codex-companion.mjs adversarial-review --wait --base origin/main`, focus text built with `--json title,body,comments` plus § Controller context. Verdict: `needs-attention`, one `[medium]` finding. **Confirmed from the branch source, and it lands on a controller ruling rather than on the worker.**

### CONFIRMED — [medium] A green suite that exercises error logging is failed, and the escape hatch cannot reach it

`failure_signature='^(FAIL|Traceback|ERROR)'` treats any column-0 `ERROR` as a failed suite. Python's `logging.error()` writes `ERROR:root:<msg>` at column 0, and `2>&1` folds stderr into what is scanned. So a passing `*_test.py` that exercises a logging path fails the merge gate.

The scan's own comment says this in full, three lines above the variable:

> Known limit: Python's `logging.error()` writes `ERROR:root:<msg>` at column 0 ... so a green `*_test.py` suite that exercises a logging path would be failed by this scan and cannot reach the indent hatch without setting its own `basicConfig(format=...)` or capturing the logger. No suite in this repo does that today.

The indent hatch is the reason I ruled against building a per-suite opt-out. That ruling rested on there being no legitimate need for one yet. **This comment is the legitimate need, written down in the same commit.** Testing an error path is ordinary, and the first person to do it meets a red gate whose message will not tell them why.

**Where I disagree with Codex, and it matters for the fix:** it says there is no recovery path. There are two, both named in that comment — `basicConfig(format=...)`, or capturing the logger, which a well-written test doing `assertLogs`/`caplog` does anyway. So this is a false positive that is recoverable but undiscoverable: the cost is a confusing red, not a wall.

That changes what the fix should be. A per-suite declaration is still the wrong shape for the reason the worker gave and I accepted — a bypass with a typo'd declaration reading as "scanning off" is defect class 1 inside the guard against class 1. The proportionate fix is that **the gate's failure output names this case and its two remedies**, so the person hitting it is one read from the answer instead of one bisect. The knowledge is already written; it is in a comment nobody reads at the moment they need it.

Round 2 carries this as a message-and-guidance change, not a mechanism change, unless the worker argues otherwise — it has overruled me twice today and been right both times.

### Controller note

The round-1 story is the reason this ticket exists. Two axes independently found that only `FAIL` had a positive fixture and only `ERROR` appeared in the narration fixture, so the signature set could be narrowed to `^(FAIL)` with the whole suite green — **while the fixture's comment asserted all three were covered.** Defect class 3 inside the guard against class 1. The verification pass then added the three one-word deletions, so every word in the set is independently load-bearing. And S2/C3/C4 were unchecked `mkdir`/`cp` in the harness, where a missing fixture would have produced a gate run over no suites, exit 0, and a red naming the wrong cause — class 1 in the test for class 1.

PR #968, opened 2026-09-21
multi-axis-code-review: one publish protocol for both capture modes; call-site mutation step (#948, #961)

### LINKED ISSUE #948: The two capture blocks in multi-axis-code-review restate one publish protocol
`multi-axis-code-review/SKILL.md` § 4 now publishes two fenced capture blocks — the `git diff` range one (grep `git rev-parse --path-format=absolute`) and the sha-list one (grep `sha-list review: the commit list is empty`). They share five lines of publish scaffolding: `n=`, `worktree=`, `tmp=$(mktemp "$dir/.diff-$n.XXXXXX")`, `mv "$tmp" "$patch"` and `wc -l "$patch"`, four of them annotated "as above". A change to the nonce scheme, the atomic-publish step, or the 14-day retention is a two-place edit, and nothing fails if only one place is edited.

Each block is extracted and run end-to-end by its own suite — `diff-capture.test.sh` and `sha-list.test.sh` — so a break inside one block is caught. Drift *between* them is not: the two suites never compare the blocks, so the range block could keep a nonce the sha-list block lost and both stay green.

Filed rather than fixed in #932 because the obvious fix — one block with a mode branch — makes the thing a caller pastes harder to read, and the tidy fix (hoist the shared scaffolding into the `$dir` preamble both modes already run) changes what `diff-capture.test.sh` extracts, so it wants its own diff and its own review rather than riding along in a three-ticket clump. Either shape is fine; the acceptance criterion is that the publish protocol is stated once and that a suite fails when the two modes disagree about it.

Filed from the standards axis (S1) of the round-1 review on PR for #932/#938/#939.

## Blocked by

- #932

### LINKED ISSUE #961: The witness check mutates the guard's body; the guard's call site is what goes untested
The correctness axis in `multi-axis-code-review/SKILL.md` (grep `strip the constraint under test`) tells a reviewer to mutate the constraint and watch the assertion go red. That catches a hollow assertion. It does not catch an unreached guard, and the unreached guard is what keeps shipping.

Shape: a diff adds a safety guard, a unit test calls the guard directly and asserts it refuses, and the guard's **call site** — the entry point the guard exists to protect — has no test that would notice if the call were removed. Mutating the guard's body reddens the unit test, so the witness passes, and the guard can still be bypassed at the only place it matters.

Live instance, #958 / PR 959, found by the Codex pass after the repo's own three-axis review missed it. `burndown/loop.py` gained `count_agent_processes`, which refuses when `ps` cannot be counted, so that an unmeasured box never reads as an empty one. `box` has a CLI test for the failed-`ps` path. `dispatch` — the command the whole safety property is about — has no CLI test that omits `--processes`, so every dispatch test supplies a count and never reaches the measurement. Codex's words: a regression that leaves `box` fail-closed but makes default `dispatch` skip `agent_count` or fall back to something permissive "would remain green". The worker had flagged the same gap honestly in its own PR-up and the axis still did not require closing it.

Build: the mutation is on the **call**, not the body. For each guard a diff adds, delete or neutralize the call to it at every entry point the guard is meant to protect, run the covering suite, and confirm it goes red at that entry point — not only in the unit test that calls the guard directly. A guard whose body-mutation reddens a test but whose call-site mutation does not is an unprotected entry point, and that is the finding. Read the failure message as always, so a red from a missing file is not mistaken for a red from the assertion.

Sequencing: `multi-axis-code-review/SKILL.md` is being edited right now on `implement-957`, which is building the concurrent witness-check machinery in that same file. This belongs with it rather than in a colliding branch — see `burndown/references/merge-tail.md`.

Acceptance: the correctness axis states the call-site mutation as a separate required step from the constraint mutation, and states what a surviving call-site mutation means. If #957's machinery runs mutations, a call-site mutation is one of the kinds it can run.

Filed from conversation — Codex adversarial pass on PR 959, 2026-09-20.

## Blocked by

None — can start immediately.

caneff:
Closes #948

Closes #961

## What changed
- § 4 of `multi-axis-code-review/SKILL.md` states the capture publish protocol once (`new_capture`, `publish_capture` in the preamble); both capture modes call it, and `diff-capture.test.sh` fails when either mode restates or drops it (#948).
- The correctness brief gains required step (e): mutate a guard's call site at every entry point, with a survivor meaning "unprotected entry point"; the witness-cost section names it as a kind of mutation the concurrent machinery runs (#961).
- Both blocks refuse a shell that never ran the preamble; suites pin it.

## Tests run
`bash multi-axis-code-review/diff-capture.test.sh`, `sha-list.test.sh`, `witness-check.test.sh` — each prints `PASS multi-axis-code-review/<name>.test.sh`. Mutations seen red on their own assertion: dropping `-$$` from the range block, removing the sha-list `publish_capture` call, removing the range block's preamble guard, renaming `unprotected entry point`.

## Decisions made
Round 1, 12 findings (ids from the sidecars; dispositions in `dispositions-948.jsonl`):
- S1 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): temp path passed by argument.
- S2 disputed: the duplicate fence extractor stays; a shared helper is a cross-suite dependency beyond #948, and both extractions now have named FAIL guards.
- S3 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): the refusal names the patch path.
- S4 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): needle bound to `mutate its call site`.
- S5 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): per-mode guards and counts (worker had called it disputed; the verifier read it as fixed).
- P1 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): nonce and file stem pinned per mode.
- P2 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): each mode counted on its own.
- C1 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): `type publish_capture` guard in both blocks, prose says same shell.
- C2 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): guard on the sha-list extraction.
- C3 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): asserts the capture landed under the scratch HOME (post hoc: the sweep has already run by then).
- C4 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): same change as S3.
- C5 fixed (659eb871c60ab6b7b1e6bd46bdb2f8154ec86575): range extraction guard no longer keyed on the call it protects.

Codex gate pass (PR comment https://github.com/caneff/agent-skills/pull/968#issuecomment-5762875972), one [high]: the final patch name was keyed on `$$`, which is per shell, so two same-input captures in one shell shared a path. Fixed (5aa74e0fd74f27cf8f37ce65d52d440fedb83f30): the suffix comes from the `mktemp` path inside the shared publish protocol; both suites gained a same-shell-twice test, seen red with the suffix removed.

Codex second pass (PR comment https://github.com/caneff/agent-skills/pull/968#issuecomment-5762956405), final round, controller dispositions:
- [high] mktemp suffix reuse after the temp file is moved — **disputed**: `mktemp` draws six characters from 62 symbols, about 5.7e10 names, at random; a released name is not handed back, it is merely eligible again, and a review directory holds tens of captures. That is the same collision bound as any random nonce, and the protocol never claimed more. A non-replacing publish (`ln` then `rm`) would make it structural; not worth a third round.
- [medium] a call-site mutation shares its id with the constraint mutation — **disputed**: the sentence means the same *kind* of id, allocated the same way; `witness-check.test.sh` refuses a duplicated id, so two mutations never share one. The wording is ambiguous enough to misread, so it is filed as #969 (prose only).

## Last reviewed sha
659eb871c60ab6b7b1e6bd46bdb2f8154ec86575 (Claude round 1). Commit 5aa74e0fd74f27cf8f37ce65d52d440fedb83f30 fixes the Codex finding and was not re-reviewed by the Claude axes.


caneff:
[codex] Thread ready (01a0c489-5b83-7872-a5fd-a827a4afdaf0).
[codex] Turn started (01a0c489-5c07-7b52-a3ca-8ff602c505a3).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’ll inspect the branch diff and execute the relevant review ...
[codex] Running command: /bin/bash -lc "sed -n '1,260p' multi-axis-code-review/SKILL.md && git status --short && git d...
[codex] Command completed: /bin/bash -lc "sed -n '1,260p' multi-axis-code-review/SKILL.md && git status --short && git d... (exit 0)
[codex] Running command: /bin/bash -lc "sed -n '200,390p' multi-axis-code-review/SKILL.md && nl -ba multi-axis-code-re...
[codex] Command completed: /bin/bash -lc "sed -n '200,390p' multi-axis-code-review/SKILL.md && nl -ba multi-axis-code-re... (exit 0)
[codex] Running command: /bin/bash -lc "wc -l multi-axis-code-review/SKILL.md && sed -n '390,780p' multi-axis-code-rev...
[codex] Command completed: /bin/bash -lc "wc -l multi-axis-code-review/SKILL.md && sed -n '390,780p' multi-axis-code-rev... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No ship: the claimed per-invocation capture isolation...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No ship: the claimed per-invocation capture isolation still fails when a caller runs two captures in the documented same shell.

Findings:
- [high] `$$` is a shell nonce, not an invocation nonce (multi-axis-code-review/SKILL.md:206-276)
  Both capture modes require the preamble to run in the same shell, but their published path is keyed by `$$`. Repeating a range capture at the same HEAD, or a sha-list capture with the same resolved list, in that shell produces the same path and `mv` replaces the earlier capture. A lagging reviewer that has received the first path but has not opened it can then read the later patch against the first round's commit/spec context—the stale-but-nonempty capture failure this protocol is meant to prevent. The current tests invoke each capture through a separate `bash` process, so their distinct PIDs mask this regression.
  Recommendation: Generate the final patch pathname from a genuinely per-invocation unique value (for example, a `mktemp`-allocated final name) in the shared publish protocol, and add an end-to-end test that performs two same-revision/same-list captures within one shell and proves the first path and contents remain unchanged.

Next steps:
- Fix the final-path nonce and add the same-shell regression test.

caneff:
[codex] Thread ready (01a0c48d-29b1-7711-84ad-0ee654454d80).
[codex] Turn started (01a0c48d-2a03-7b20-9c66-c2863f1ad7fb).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’m inspecting the branch diff and its executable capture/wit...
[codex] Running command: /bin/bash -lc "wc -l /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md; sed -n '241...
[codex] Command completed: /bin/bash -lc "wc -l /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md; sed -n '241... (exit 0)
[codex] Running command: /bin/bash -lc "sed -n '521,620p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md;...
[codex] Command completed: /bin/bash -lc "sed -n '521,620p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md;... (exit 0)
[codex] Running command: /bin/bash -lc 'git show --format=fuller --stat aafcb66 && git show --format=fuller --stat 102...
[codex] Command completed: /bin/bash -lc 'git show --format=fuller --stat aafcb66 && git show --format=fuller --stat 102... (exit 0)
[codex] Running command: /bin/bash -lc "nl -ba multi-axis-code-review/SKILL.md | sed -n '300,455p'; git status --short...
[codex] Command completed: /bin/bash -lc "nl -ba multi-axis-code-review/SKILL.md | sed -n '300,455p'; git status --short... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: the new publish name can still be reused and...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new publish name can still be reused and overwrite a live capture; the #961 mutation recipe also cannot represent its required separate call-site checks.

Findings:
- [high] Published capture name is not collision-safe after its temp file is moved (multi-axis-code-review/SKILL.md:190-196)
  `mktemp` only reserves the temporary pathname while that file exists. After the first invocation moves it away, a later invocation can obtain the same `.diff-$n.<suffix>` temporary name; `publish_capture` derives the same final name and plain `mv` overwrites the prior capture. A lagging reviewer can therefore read a replaced-but-non-empty patch—the exact stale/wrong-diff failure this protocol is meant to prevent. The current tests exercise different random suffixes, not reuse of a released temporary name.
  Recommendation: Publish to a destination that is atomically created without replacement (and fail/retry on an existing destination), rather than deriving an overwrite-permitted destination from a temporary filename. Add a deterministic collision test that forces a second capture to reuse a prior temp suffix and verifies the first final patch remains unchanged.
- [medium] Call-site mutations are collapsed into test IDs instead of being independently runnable (multi-axis-code-review/SKILL.md:341-343)
  The correctness brief requires a separate mutation for every protected entry point, but the machinery says a call-site mutation uses the same ID as the constraint mutation while the input admits only one ID per new/changed test and rejects duplicate IDs. A newly added guard with a direct unit test plus two protected entry points consequently has no way to run and report three distinct mutations/worktrees. Combining them under one ID can let one red suite mask an untested or surviving call site, defeating the #961 acceptance criterion.
  Recommendation: Define distinct mutation IDs for each constraint mutation and each guard-entrypoint call-site mutation, update the recipe's enumeration contract accordingly, and add a witness test covering a guard with multiple entry points.

Next steps:
- Make publication non-overwriting under suffix reuse and add a forced-collision test.
- Give call-site mutations unique IDs per protected entry point.

Issue #991, opened 2026-09-21
Share brief, registry and herdr code between worker alert hooks


caneff:
`flow/claude/hooks/worker-spin-alert.sh` copies four pieces from `worker-stop-alert.sh`: the `/implement` brief-capture regex, the `~/.claude/sessions/*.json` registry loop with its `/proc/<pid>/stat` field-22 liveness check, `logline()`, and the herdr agent-name and pane lookup. Extract one helper both hooks source, so a registry schema change fixes both; today a fix to one leaves the other resolving no controller and exiting 0 silently. Both `*.test.sh` files must stay green and each hook keeps its own alert text and dedupe key.

Filed from the #925 review (standards finding S2).

## Blocked by

- #925

PR #992, opened 2026-09-21
Alert the controller when a worker repeats one tool call (#925)

### LINKED ISSUE #925: Liveness: a worker spinning on identical no-op tool calls reads as healthy

caneff:
Closes #925

## What changed
- `flow/claude/hooks/worker-spin-alert.sh`: a PostToolUse hook that types an alert into the controller's herdr pane when a worker makes 20 identical consecutive tool calls; `--classify <transcript>` runs the same check on demand.
- Wired in `settings.json`, `install.sh` (link) and `install.test.sh`; documented in `OPERATIONS.md`.
- Tests: `worker-spin-alert.test.sh` with five transcript fixtures plus in-test padded, no-brief and not-sent cases.

## Tests run
- `bash tests/all.sh` → `73 suites passed`
- `bash flow/claude/hooks/worker-spin-alert.test.sh` → `ALL PASS` (11 checks)

## Decisions made
- Fires from a PostToolUse hook, not Stop: a spinning worker never stops. Stated on #925 in a comment. N=20 (`SPIN_N` overrides).
- Files beyond the hook and its test, all wiring the ticket needs to be reachable: `settings.json`, `install.sh`, `install.test.sh`, `OPERATIONS.md`.
- Round 1 findings, with dispositions (sidecar `dispositions-925.jsonl`):
  - S1 fixed 49c0a4f (`--classify` refuses an unreadable path)
  - S2 filed #991 (shared helper with `worker-stop-alert.sh`)
  - S3 disputed: the ticket asks for transcript fixtures; committed files are readable and match the incident
  - S4 fixed 49c0a4f (only a `sent` line dedupes, so an unsent alert retries on the next call; no backoff)
  - P1 fixed: decision comment posted on #925
  - P2 disputed: the `--controller` gate mirrors `worker-stop-alert.sh`, which the ticket asked to mirror
  - C1 fixed 49c0a4f (window counts tool-call lines, not raw lines)
  - C2 fixed 49c0a4f (stub newline; hollow dedupe witness)
  - C3 fixed 49c0a4f (padded fixture with sidechain calls)
  - C4 fixed 90f034f (no-brief case also asserts no log line)
  - C5 fixed 49c0a4f (same fix as S1)
  - C6 fixed 49c0a4f (same fix as S4)
- Codex gate pass (final round), both classified `codex-only`: (1) [high] "structural jq equality, not byte-identical input" disputed, controller ruling: the transcript stores the harness's serialization of the input, not the model's bytes, so structural equality of the stored object is the contract; the hook header now says so. (2) [medium] alert key truncated the input at 120 chars, suppressing a second spin with a shared prefix: fixed in 3599b64 (key is a digest of the full input; only the alert text truncates; a same-prefix fixture test covers it).
- `count` is a floor (last 500 tool-call lines).

Codex second pass (final round, controller dispositions):
- [high] a later spin of the same call after an interrupting call is deduplicated for the rest of the session — **filed** as #998. Real and narrow; the once-per-streak identity needs a run boundary.
- [medium] structural equality rather than byte-identical input — **disputed**, as at the gate: the transcript stores the harness's serialized object, not the model's bytes, so structural equality on the stored input is the implementable contract, and the header comment now says so.

## Last reviewed sha
90f034f by the Claude axes; the Codex fix commit after it was not re-reviewed.


caneff:
[codex] Thread ready (01a0c564-26fd-7de3-b819-c4d3d5195375).
[codex] Turn started (01a0c564-2830-7263-92e0-fb39bd0f2146).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’ll inspect the branch diff and exercise the classifier/test...
[codex] Running command: /bin/bash -lc "sed -n '241,520p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md ...
[codex] Command failed: /bin/bash -lc "sed -n '241,520p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md ... (exit 1)
[codex] Assistant message captured: {"verdict":"approve","summary":"The mandatory multi-axis review skill applies here; I’m using...
[codex] Running command: /bin/bash -lc "nl -ba flow/claude/hooks/worker-spin-alert.sh | sed -n '20,100p' printf '%s\\n...
[codex] Command completed: /bin/bash -lc "nl -ba flow/claude/hooks/worker-spin-alert.sh | sed -n '20,100p' printf '%s\\n... (exit 0)
[codex] Running command: /bin/bash -lc 'bash flow/claude/hooks/worker-spin-alert.sh --classify <(for i in $(seq 1 20);...
[codex] Command completed: /bin/bash -lc 'bash flow/claude/hooks/worker-spin-alert.sh --classify <(for i in $(seq 1 20);... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: the classifier violates the ticket’s byte-id...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the classifier violates the ticket’s byte-identical invariant, and alert deduplication can hide later distinct spins.

Findings:
- [high] Classifier treats differently serialized inputs as identical (flow/claude/hooks/worker-spin-alert.sh:31-39)
  The classifier parses inputs into jq objects and compares them with `==`. jq object equality ignores key order, so 20 calls whose raw inputs alternate between `{"command":"echo ok","timeout":1}` and `{"timeout":1,"command":"echo ok"}` are flagged; I reproduced this with `--classify`. Those inputs are not byte-identical, so this violates the stated classifier contract and can alert on varied calls.
  Recommendation: Preserve and compare the serialized input bytes (or explicitly revise the requirement to structural equality). Add a fixture alternating key order/whitespace that must not spin.
- [medium] Truncated input makes distinct spin runs share a dedupe key (flow/claude/hooks/worker-spin-alert.sh:65-72)
  `input` is truncated to 120 characters before it becomes part of `key`. Two distinct long inputs with the same first 120 serialized bytes, each repeated N times in the same session, therefore share the first run’s sent record; the second alert is suppressed. This defeats the documented one-alert-per-run behavior and hides a new stuck loop.
  Recommendation: Use the complete input, or a collision-resistant digest of it, in the dedupe key; truncate only the human-facing alert text. Add a hook test for two long, same-prefix repeated runs.

Next steps:
- Fix raw-input comparison and dedupe identity, then add the corresponding transcript fixtures.

caneff:
[codex] Thread ready (01a0c567-1268-7180-a516-a32472eed730).
[codex] Turn started (01a0c567-12f8-75b2-8f3a-81ff4d033887).
[codex] Assistant message captured: {"verdict":"approve","summary":"I’m inspecting the branch diff and exercising the classifier ...
[codex] Running command: /bin/bash -lc "sed -n '1,260p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md &&...
[codex] Command completed: /bin/bash -lc "sed -n '1,260p' /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md &&... (exit 0)
[codex] Running command: /bin/bash -lc "wc -l /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md && sed -n '2...
[codex] Command completed: /bin/bash -lc "wc -l /home/caneff/.agents/skills/multi-axis-code-review/SKILL.md && sed -n '2... (exit 0)
[codex] Assistant message captured: {"verdict":"approve","summary":"Using the repository’s multi-axis review procedure because th...
[codex] Running command: /bin/bash -lc "sed -n '110,180p' flow/claude/settings.json; sed -n '1,220p' flow/claude/hooks...
[codex] Command failed: /bin/bash -lc "sed -n '110,180p' flow/claude/settings.json; sed -n '1,220p' flow/claude/hooks... (exit 1)
[codex] Running command: /bin/bash -lc "nl -ba flow/claude/hooks/worker-spin-alert.sh | sed -n '20,110p'; git diff ori...
[codex] Command completed: /bin/bash -lc "nl -ba flow/claude/hooks/worker-spin-alert.sh | sed -n '20,110p'; git diff ori... (exit 0)
[codex] Assistant message captured: {"verdict":"needs-attention","summary":"No-ship: the hook can permanently suppress a later, i...
[codex] Turn completed.
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the hook can permanently suppress a later, independent spin in the same worker session, and its classifier violates the required byte-identical-input boundary.

Findings:
- [high] A later spin with the same call is silently deduplicated (flow/claude/hooks/worker-spin-alert.sh:75-79)
  The dedupe key contains only session, tool, and input digest. Once one run is logged as sent, a different tool call can interrupt it and a later run of the same call can again reach N, but it matches the old sent record and exits without alerting. This contradicts the documented “once per run” behavior and leaves a resumed worker spin invisible for the rest of its session.
  Recommendation: Include a run-specific boundary in the dedupe identity (for example, the first tool-use ID of the current consecutive streak), and add a fixture/test for spin → different tool → same spin that expects two alerts.
- [medium] Structural comparison flags inputs that are not byte-identical (flow/claude/hooks/worker-spin-alert.sh:36-44)
  The classifier parses inputs with jq and compares objects using jq equality. Object key order is therefore ignored, despite the ticket and operations document requiring byte-identical input. A worker alternating otherwise equivalent inputs whose serialized key order differs can be classified as one repeated run and falsely alert; this also violates the requirement that varied work never raises the alert.
  Recommendation: Preserve and compare the serialized input bytes (or an unambiguous digest of those bytes) before JSON object normalization, and add a transcript fixture where equivalent inputs have different key order and must not spin.

Next steps:
- Fix the run-boundary dedupe and byte-level comparison, then add the two regression fixtures.

Issue #1017, opened 2026-09-22
notify-send hangs when a value flag is the last argument


caneff:
`flow/bin/notify-send` (moved verbatim by #874, so the bug predates it): a value-taking flag (`-a -i -u -t -c -h -A -r` and their long forms) as the final argument makes `shift 2` fail with one argument left, nothing shifts, and the `while [ $# -gt 0 ]` loop spins forever, so the shim hangs. Real `notify-send` errors instead. The argument parser has no test; add one that runs each such flag last, under a timeout, and fix the shift.

Filed from: #874 round 1 review, finding C2.

## Blocked by

- #874

Issue #1016, opened 2026-09-22
Call herdr-toast-install from flow/install.sh


caneff:
`flow/install.sh` is the repo's single install mechanism, but it never calls `flow/bin/herdr-toast-install`, so a fresh-machine install leaves the herdr toast scripts unlinked and the `herdrfocus` handler unregistered. Route the toast step from `install.sh`. The installer refuses a linked worktree and any path under /tmp, so `install.sh` must handle that refusal without aborting the rest of the install.

Filed from: #874 round 1 review, finding S1.

## Blocked by

- #874

PR #1018, opened 2026-09-22
Move the herdr toast scripts into flow/bin (#874)

### LINKED ISSUE #874: Track the herdr toast scripts in flow/bin, symlinked from ~/.local/bin
The five scripts behind herdr's `[ui.toast] delivery = "system"` live only in `~/.local/bin` and are in no git repo: `notify-send` (the WSL shim that raises a Windows toast), `run-hidden.vbs`, `herdr-focus.vbs` (the `herdrfocus:` URL handler), `herdr-focus-pick.ps1` (raises the host window by owning process) and `herdr-focus-latest` (focuses the pane). Two sessions edited them on 2026-09-17 with only `.bak-*` copies as a safety net. `job-run` shows the pattern to follow: source in `flow/bin/`, `~/.local/bin/<name>` a symlink to it (`ls -la ~/.local/bin/job-run`).

Move the five into `flow/bin/` and replace each `~/.local/bin` copy with a symlink. Watch two things: (1) `herdr-focus.vbs` and the `HKCU\Software\Classes\herdrfocus\shell\open\command` registry value reference the scripts by `\\wsl.localhost\Ubuntu-24.04\home\caneff\.local\bin\...` UNC paths, and `notify-send` references `run-hidden.vbs` the same way — a Windows-side reader may not follow a WSL symlink, so either keep the UNC targets pointing at the real files or confirm wscript/powershell resolve the link; (2) the `.bak-*` files stay out of the repo. Chain and test method: `flow/claude/OPERATIONS.md` § Herdr configuration, "Toast clicks follow the host". Check: `HERDR_TOAST_PANE=<pane id> notify-send "t" "b"` from another workspace still raises Zed and jumps to that pane after the move.

Filed from: conversation (herdr toast fix after the Zed move, 2026-09-18).

## Blocked by

None — can start immediately.

caneff:
Closes #874

## What changed
- The five herdr toast scripts move into `flow/bin/`; the two Windows-read paths (`notify-send` → `run-hidden.vbs`, `herdr-focus.vbs` → `herdr-focus-pick.ps1`) now derive from the script's own real location, since Windows cannot follow a WSL symlink over `\\wsl.localhost` (checked: `Get-Content` on `~/.local/bin/job-run` fails).
- New `flow/bin/herdr-toast-install` links them into `~/.local/bin` and points the `herdrfocus` registry handler at `flow/bin`; it refuses a linked worktree or any `/tmp` path. `OPERATIONS.md` updated.
- **Not yet applied to the machine:** the `~/.local/bin` symlinks and registry value change only when the installer runs from the primary checkout after merge. Until then the ticket's end-to-end check (`HERDR_TOAST_PANE=<pane> notify-send "t" "b"`) exercises the old copies, so it proves nothing about the move; run it after the installer.

## Tests run
`bash flow/bin/herdr-toast-install.test.sh` → `ok` (fixture HOME, `powershell.exe` stub first on PATH, asserts the stub was called exactly once with the fixture path).

## Decisions made
Round 1 (13 findings, ids per sidecar; dispositions in `~/.cache/agent-reviews/skills/dispositions-874.jsonl`):
- S1 filed #1016 (call the installer from `flow/install.sh`)
- S2 fixed 8a02050 (real file kept as `.pre-flow`)
- S3 fixed 8a02050 (stale header comment)
- S4 fixed 8a02050 (install root derived, not spelled three times)
- S5 fixed 8a02050 (test covers all four Windows-read files)
- S6 fixed 8a02050 (test executable)
- P1 disputed: the end-to-end toast check is outstanding until the installer has run; stated above and in `OPERATIONS.md`
- P2 disputed: `herdr-focus-latest` stays under `~/.local/bin`, which the installer creates; it runs in WSL, so the UNC limit does not apply
- C1 fixed 8a02050 (PATH stub, exactly-once assertion; skip env var removed)
- C2 filed #1017 (pre-existing `notify-send` hang on a trailing value flag, moved verbatim)
- C3 fixed 8a02050 (refuses worktree and `/tmp`, per controller ruling)
- C4 fixed 8a02050
- C5 fixed 8a02050 (refuses a real directory)

Verification pass confirmed all; it added V1 (no test executes `notify-send`'s new bindir derivation; left, low) and V2 (stub covers only the bare `powershell.exe` call), addressed by a comment in 6901481.
Incident: during round 1 a correctness mutation ran the registry step for real; the value was restored by Chris.

## Last reviewed sha
8a02050. Commit 6901481 (a test-header comment) was not re-reviewed.

Gate pass on 6901481 (needs-attention, three findings):
- [high] shims replaced before the registry write: **fixed** af5a09f — registry first, then links; failing-registry test added.
- [medium] unquoted `.vbs` path in the registered command: **fixed** af5a09f — path and `%1` quoted, exact-argument assertion, fixture with a space; `reg.exe` called directly.
- [medium] `.pre-flow` backup overwritten on retry: **disputed** — a retry skips already-linked names and unreached names have no backup yet; only a hand-copied restore reaches the branch, with identical content.

Second pass on af5a09f (needs-attention, one finding):
- [medium] an `ln -sfn` failure after `mv` in the same directory leaves `herdr-focus-latest` absent: **disputed** — the symlink write after a successful rename in the same local directory is not a failure the installer is asked to survive, and the script is re-runnable (the moved file is restored by hand from `.pre-flow`, then re-run). Rollback machinery is out of this ticket's scope. No third run: controller disposition.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the installer can leave the live toast path broken on a routine registry/tool failure, and emits an unquoted Windows handler path.

Findings:
- [high] Installer mutates live shims before the registry update can fail (flow/bin/herdr-toast-install:28-38)
  The script replaces all five ~/.local/bin files before invoking wslpath and reg add. If either Windows-side command fails, set -e exits after the links are live but before the handler points at flow/bin. An existing handler that still targets ~/.local/bin then reaches a WSL symlink that Windows cannot follow, breaking toast clicks with no rollback. The test only covers a successful registry stub call.
  Recommendation: Make installation transactional: preflight required tools, preserve the prior handler and destinations, and restore them on any failure; alternatively perform the registry update only after a rollback mechanism is armed and test its failing path.
- [medium] Registry command fails for a checkout path containing spaces (flow/bin/herdr-toast-install:36-38)
  The dynamically derived UNC script path is written into the URL-handler command without quotes. For a valid primary checkout whose path contains a space, Windows splits the .vbs path into arguments and wscript cannot open the handler, so every toast click fails. The test checks only that the path substring is present, not that the resulting command line is executable.
  Recommendation: Quote the .vbs path (and the %1 argument) in the registered command, with escaping appropriate for the PowerShell/reg invocation; add a fixture path containing spaces and assert the exact registered command.
- [medium] Existing recovery backups are silently overwritten (flow/bin/herdr-toast-install:29-31)
  When a real destination file exists, mv unconditionally uses the fixed <name>.pre-flow target. A retry after a partial install or manual restoration overwrites the prior recovery copy, destroying the only preserved version of the earlier live script. This is particularly risky because the change is replacing previously untracked, manually maintained files.
  Recommendation: Refuse when <name>.pre-flow already exists, or create a collision-free backup name and report it before replacing the destination; cover the collision case in the installer test.

Next steps:
- Fix the transactional failure and quoting paths, then run the installer test with registry/wslpath failures and a flow path containing spaces.

caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: a filesystem failure after the registry write can leave toast clicks permanently unable to focus panes until manual repair.

Findings:
- [medium] Post-registry link failure leaves the active handler with no focus target (flow/bin/herdr-toast-install:35-43)
  The registry is updated before any links are changed, then each existing file is moved before its replacement symlink is created. If the final `ln` fails after `herdr-focus-latest` has been moved (for example, an I/O or filesystem error), the new registry handler runs `flow/bin/herdr-focus.vbs`, which still invokes `~/.local/bin/herdr-focus-latest`; that path is absent and the click cannot focus its pane. `set -e` exits without restoring either the moved file or the previous registry value. The tests cover registry failure before mutations, but not a failure between `mv` and `ln`.
  Recommendation: Make the filesystem and registry transition recoverable: record prior registry/filesystem state and restore it on any post-registry failure, or otherwise arrange the handler so it does not depend on a partially installed local link. Add a test that forces `ln` to fail after moving `herdr-focus-latest` and verifies clicks retain a working handler/target.

Next steps:
- Add rollback or a failure-safe installation sequence, then exercise the injected link-failure case.

Issue #1042, opened 2026-09-22
controller-restore's main() has no integration test


caneff:
Every sibling lane binary (`resolve-controller`, `merge-cleanup`, `implement-dispatch`) has a `tests/<binary>.rs` integration test exercising the real binary end to end against the fake `gh`/`herdr` from `lane-fake`; `controller-restore` (added in #964) only has unit tests for its pure helper functions (`restore_line`, `describe_pr`, `parse_pr_status`, `is_subagent`, `agent_status`) — its `main()`, the `is_subagent` stdin-JSON gate, `find_own_pid` resolution, and the actual `herdr`/`gh` subprocess invocations are never exercised by the suite.

Doing this properly needs `lane-fake`'s `gh pr list` fake extended to answer with a caller-chosen `<number> <STATE>` pair (today it always answers `<number> <oid>`, tailored to `merge-cleanup`'s own jq shape) so a test fixture can exercise the OPEN/MERGED/CLOSED/no-PR branches through the real binary.

Raised by #964's spec axis (P1). A manual real-subprocess smoke test was run and recorded in #964's PR body in the meantime, but that is not a standing regression test.

Filed from #964's multi-axis review, 2026-09-22.

## Blocked by

- #964

Issue #1041, opened 2026-09-22
controller-restore can print a stale worker record after pid reuse


caneff:
`flow/lane/src/bin/controller_restore.rs` finds its own session's pid via `sessions::find_own_pid`, which validates the live session's own `<pid>.json` against `/proc/<pid>/stat`'s starttime — but the sibling `<pid>.workers.jsonl` it then reads has no equivalent validation. If a Claude session at pid X dispatches workers, exits without those workspaces ever being cleaned up via `merge-cleanup`, and pid X is later reused by an unrelated new Claude session, that new session's `controller-restore` hook reads and prints the old, stale worker records as if they were its own. `dispatched_at` is recorded in each record but nothing ever compares it against anything.

Raised by #964's correctness axis (C3), rated PLAUSIBLE.

Fix direction: store the dispatching session's own sessionId (or `procStart`) in each `WorkerRecord` and validate it against the resuming session's own live sessionId before printing a restore line for it.

Filed from #964's multi-axis review, 2026-09-22.

## Blocked by

- #964

caneff:
Fixed before this was dispatched: PR #1043's Codex fix round (66e16cb, landed d578e5c) added proc_start to WorkerRecord and drop_stale in controller_restore.rs, with a unit test. Closed by the burn controller.

Issue #1040, opened 2026-09-22
Worker-record cleanup match is symlink-fragile (exact path string)


caneff:
`flow/lane/src/workers.rs::remove_workspace` matches a worker record's `workspace` field (built in `implement_dispatch.rs` from `wt.display()`) against the workspace path `merge-cleanup` passes it, by exact string equality — two independent construction sites, no shared normalizer. A symlinked repo path, or any path spelled two different but equivalent ways between the two binaries, makes the record survive cleanup silently (`remove_workspace` returns `false`, nothing is printed), and a later `/clear` on that controller hands out a cleanup line for a workspace that's already gone.

Raised independently by #964's standards axis (S5), spec axis (P4) and correctness axis (C5) reviews.

Fix direction: canonicalize both sides (`std::fs::canonicalize` or a shared helper) before compare, or store/compare on the worktree's git-reported absolute path rather than a `display()`-derived string.

Filed from #964's multi-axis review, 2026-09-22.

## Blocked by

- #964

PR #1043, opened 2026-09-22
Record controller/worker pairs at dispatch, restore on /clear

### LINKED ISSUE #964: implement-dispatch records controller→worker pairs; a SessionStart hook restores them after /clear
Chris asked (2026-09-21) for the controller lane to be robust to `/clear`. A clear wipes the context but not the process, so the session name in `~/.claude/sessions/<pid>.json` survives and the worker's messages still arrive; what is lost is the knowledge of which workers this session controls, plus any `notify_when_idle` subscriptions, which are per context. A "PR up" that lands during the clear, or an idle notice for a merged PR awaiting cleanup, is then silently dropped.

Change, two halves:

1. **Record the pair at dispatch.** `implement-dispatch` (the script behind `implement/SKILL.md` § Dispatch) appends one line per dispatched worker to the controller's session record, e.g. a `workers` array in `~/.claude/sessions/<pid>.json` or a sibling `<pid>.workers` file: `{agent, ticket(s), branch, workspace, repo, cleanup line, chris_merges, dispatched_at}`. `merge-cleanup` removes the entry when it removes the workspace.
2. **Restore on session start.** A SessionStart hook (it fires on `/clear` too; `~/.claude/settings.json` already has a `SessionStart` array) reads that record for the current process, asks herdr (`herdr agent get <name>`) which agents are still alive and their status, checks `gh pr list --head <branch>` for an open or merged PR, and prints one line per worker: `You control implement-143 (sudokupad-art-143): PR #152 open, not merged — follow implement/SKILL.md § The merge; cleanup: <line>`. Re-arm nothing automatically; the printed state is enough for the context to act.

Keep it read-only apart from the record file; never re-send a brief. Document in `flow/claude/OPERATIONS.md` § Control.

Done means: dispatch writes the record; a manual `/clear` in a session with a live worker prints the restore line; `merge-cleanup` clears the entry; existing dispatch tests pass.

Filed from conversation with Chris, sudokupad-art controller session, 2026-09-21.

## Blocked by

caneff:
Closes #964

## What changed

- `implement-dispatch` appends a `WorkerRecord` (`agent`, `tickets`, `branch`, `workspace`, `repo`, `cleanup`, `chris_merges`, `dispatched_at`) to a `<pid>.workers.jsonl` sidecar next to the controller's own `~/.claude/sessions/<pid>.json`; `merge-cleanup` removes the record for a workspace once it tears it down.
- A new `SessionStart` hook binary, `controller-restore`, reads that sidecar for the resuming session's own pid (fires on `/clear` too), asks `herdr agent list` which of those workers' agents are still alive and `gh pr list` whether each branch has an open/merged/closed PR, and prints one restore line per worker — read-only besides the record file, never re-sends a brief, re-arms nothing.
- `flow/claude/settings.json`'s `SessionStart` array now runs `controller-restore`; `flow/claude/OPERATIONS.md` § Control documents the mechanism.

## Tests run

`cd flow/lane && cargo test` → 266 passed, 0 failed, 0 ignored, across all targets (lib, `controller-restore`, `implement-dispatch`, `merge-cleanup`, `resolve-controller`, `lane-fake`, and their integration test files). Existing `tests/implement_dispatch.rs` suite: 82/82 passed, unaffected.

## Decisions made

- Used `herdr agent list` (already typed/tested via `lane::herdr::parse_agents`) instead of the ticket's literal `herdr agent get <name>` suggestion, for consistency with every other binary in this crate.
- Recorded worker records in a sibling `<pid>.workers.jsonl` file — one of the two shapes the ticket explicitly offered — rather than a `workers` array embedded in the session JSON.
- `WorkerRecord.repo` is the `owner/name` slug (what `gh pr list --repo` takes), not the absolute checkout path (already embedded in `cleanup`).
- No new crate dependency for the ISO8601 timestamp — shells out to `date -u +%Y-%m-%dT%H:%M:%SZ`, best-effort, empty string on failure — consistent with the hard rule to ask before adding a dependency.
- `controller-restore` is silent (no output, exit 0) whenever there is nothing to restore or its own session record can't be resolved, since it is a non-blocking `SessionStart` hook.

### Round-1 findings (multi-axis review, head `2e37fbc`)

- **S1** (hard, standards) — herdr/gh being unreachable printed identically to a real "no live agent"/"no PR" answer. **Fixed**, `003b3c1`.
- **S2** (judgement, standards) — `workers::read` silently drops a malformed line, no count. **Disputed**: matches the file's own documented skip-malformed-line intent; a hook must stay silent-by-default.
- **S3** (judgement, standards) — `remove_workspace` collapses unreadable-dir/file/write-failure into one `false`. **Disputed**: best-effort bookkeeping alongside the workspace's own authoritative removal.
- **S4** (judgement, standards) — `WorkerRecord::tickets`/`dispatched_at` read as yagni. **Disputed**: both are literally named in #964's own field list.
- **S5** (judgement, standards) — workspace-path exact-string match is symlink-fragile. **Filed**, caneff/agent-skills#1040.
- **P1** (judgement, spec) — `controller-restore`'s `main()` has no integration test. **Filed**, caneff/agent-skills#1042.
- **P2** (judgement, spec) — restore-line text deviates from the ticket's literal example (adds an agent-status clause and a `, Chris merges` suffix). **Disputed**: both derive from spec-required fields/behavior, intent-preserving.
- **P3** (judgement, spec) — `gh pr list --state all` is a superset of "open or merged". **Disputed**: accepted, worker-declared, superset is honestly reported.
- **P4** (judgement, spec) — same path exact-match issue as S5/C5. **Filed**, caneff/agent-skills#1040.
- **C1** (CONFIRMED hard, correctness) — `gh pr list`'s `.[0]` on an empty result interpolates as literal `null null`, read as a real PR. **Fixed**, `003b3c1`.
- **C2** (CONFIRMED class 1, correctness) — same root cause as S1. **Fixed**, `003b3c1`.
- **C3** (PLAUSIBLE, correctness) — a controller session's pid can be reused; the sidecar has no procStart/sessionId validation. **Filed**, caneff/agent-skills#1041.
- **C4** (PLAUSIBLE, correctness) — `workers::append`'s `writeln!` was two separate `write(2)` calls, risking interleaved corruption between racing dispatches. **Fixed**, `003b3c1`.
- **C5** (PLAUSIBLE, correctness) — same path exact-match issue as S5/P4. **Filed**, caneff/agent-skills#1040.
- **C6** (correctness) — `is_subagent` only checked `agent_id`, not `agentId` (camelCase), unlike `worker-stop-alert.sh`. **Fixed**, `003b3c1`.
- **C7** (correctness) — the `sessions.rs` refactor risked a behavior change in `name_of_session` when two live records share a sessionId. **Fixed**, `003b3c1` (reverted to the pre-#964 single-pass scan).

Verification pass (scoped to round-1 findings + fix commit `003b3c1`) confirmed all 16 dispositions independently, re-derived the `gh` `null null` bug live, witnessed three mutations red for the right reason, and confirmed `cargo test` at 266/266.

## Last reviewed sha

`003b3c199120` — no commits past it; this is HEAD.

Gate pass on 003b3c1 (needs-attention, three findings), all **fixed** in 66e16cb by controller ruling:
- [high] cleanup's read/filter/rewrite raced dispatch's append: one exclusive `File::lock()` on the sidecar across the read-modify-write, with an append-during-removal test.
- [high] records keyed by pid alone restored into a reused pid after a restart: `WorkerRecord` carries the session's `proc_start`; restore drops mismatches and reports the count.
- [medium] 15 s hook timeout under a 10 s + 10 s-per-worker query budget: one 12 s deadline shared by every query; a worker whose query never ran prints `unchecked`.

Second pass on 66e16cb (needs-attention, one finding):
- [medium] `read` does not take the sidecar lock, so a `/clear` during cleanup's truncate-rewrite window restores nothing: **filed** as #1044. No third run: controller disposition.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: record removal can lose live workers, PID reuse can restore another session’s workers, and the hook deadline cannot cover its serialized queries.

Findings:
- [high] Cleanup can erase a concurrently appended worker record (flow/lane/src/workers.rs:105-125)
  `remove_workspace` reads an entire JSONL file, filters it, then overwrites that same path. An `implement-dispatch` append after the read but before `write` is silently overwritten, even when it names a different workspace/controller. The next `/clear` omits that live worker with no recovery signal.
  Recommendation: Serialize append and removal with one per-record-file lock, and perform the filter/rewrite while holding it. Add a deterministic append-during-removal test.
- [high] Worker records are not bound to a unique session lifetime (flow/lane/src/workers.rs:64-72)
  The sidecar name is keyed only by PID. Session lookup validates `procStart`, but the worker file does not store or validate it. If an old controller exits without cleanup and its PID is reused, a new session's valid registry record resolves to the same sidecar and receives stale workers as if it controlled them.
  Recommendation: Bind records to sessionId or procStart (in both filename and/or record), and verify that binding before restore; discard or isolate mismatched stale records.
- [medium] The 15-second hook timeout is shorter than the serialized query budget (flow/claude/settings.json:191-193)
  The hook has a 15-second timeout, while restore first permits `herdr agent list` up to 10 seconds and then permits up to 10 seconds for each worker's `gh pr list`. A slow herdr or gh call therefore kills the hook before all restore lines are emitted; with multiple records this is guaranteed under ordinary degraded dependency behavior. `|| true` makes this loss quiet.
  Recommendation: Use a total deadline that fits the configured hook timeout and emit bounded per-worker results, or raise the hook timeout enough for the maximum supported record count; add an integration test with delayed command responses.

Next steps:
- Fix the record mutation race and session-lifetime binding, then test restore under timeout/degraded-command conditions.

caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new rewrite lock still permits restore to read a transient empty sidecar and silently drop unrelated controlled workers during /clear.

Findings:
- [medium] Restore reads a torn cleanup rewrite without taking the file lock (flow/lane/src/workers.rs:93-98)
  `remove_workspace` truncates then rewrites a controller’s whole JSONL file while holding an exclusive advisory lock, but `read` uses an unlocked `read_to_string`. With two records, cleanup of A can execute `set_len(0)`; a SessionStart caused by `/clear` then reads an empty file before B is rewritten and exits with no restore line. This defeats the core recovery path precisely when a PR/idle notification can be lost. The existing race test covers append versus cleanup, not restore versus cleanup.
  Recommendation: Make `read` participate in the same lock protocol (or use an atomic replacement design with a stable lock) so it only observes a complete generation. Add a concurrent test that pauses cleanup after truncation and proves restore reads the retained record rather than an empty file.

Next steps:
- Fix the reader/rewrite race and add the regression test; the full Cargo suite could not be run because this review environment is read-only.

Issue #1049, opened 2026-09-22
loop.py dispatch: a same-tick collision past the free-slot cut gets no held line


caneff:
`picks` (`burndown/loop.py`) breaks its loop at `len(picked) == free`, before it walks the remaining `dispatchable` clumps. A same-tick collision among clumps past that cutoff is therefore never examined and never reported — only a collision among clumps actually walked before the cut gets a `held` line.

`burndown/SKILL.md` § The loop step 5 says dispatch "names what holds the rest," which this only partially satisfies: at a low `--free` relative to a family's size, a controller sees fewer picks and idle slots with no `held` line explaining a same-tick collision that never got examined.

Evidence: `candidates_781()` in `burndown/loop_test.py` — a hub family of three (452, 457, 458 sharing one file) plus one independent clump (501) — with `--free 1`, `picks` picks #452 and stops; #457 and #458, which would also collide with #452, are neither picked nor named as held.

Filed from: the spec axis review of #971/#988/#1003 (finding P3), 2026-09-22.

## Blocked by

- None — can start immediately.

Issue #970, opened 2026-09-21
burndown: a collision component is a family run serially, not one clump


caneff:
`burndown/closure.py` returns each connected component of the file-collision graph as one clump, and `CONTEXT.md`'s Clump entry defines it that way: one worker, one workspace, one PR. On run `burn-2026-09-21-0930` (agent-skills, 2026-09-21) the 27 remaining `ready-for-agent` tickets chained through hub files (`implement/SKILL.md`, `flow/lane/src/bin/merge_cleanup.rs`, `burndown/loop.py`, `AGENTS.md`, `tests/all.sh`) into one component of 34 files, so the loop's only move was one worker holding 27 tickets in one context, one branch and one PR. Chris's ruling, 2026-09-21: "we should account for this in burndown skill automatically".

What the component actually proves is only that its tickets cannot run **concurrently**, not that they must ship together. Split the two ideas:

- A **family** is a connected component of the collision graph. Its members never hold live workspaces at the same time.
- A **clump** inside a family is the maximal set of tickets whose closures are **identical** (the same file set), capped at three tickets; any other family member is a clump of one. Identical closures would only rebase onto each other for nothing; different closures that merely overlap serialize instead.
- Serialization needs no new mechanism: `loop.py dispatch` already holds a clump whose closure intersects a live workspace's, so dispatching the lowest free clump of a family and letting the hold keep the rest is the loop as written. The candidate list `dispatch` reads is now the family's clumps, not the family.

Change `closure.py`'s output and `render` to print families and their clumps, keep the announcement line, and emit the clump list in the JSON shape `loop.py read_clumps` takes (today `closure.py` prints text and the candidates file is built by hand — close that gap here, one `--json` flag or a second render). Update `burndown/SKILL.md` § Clumping and § The loop step 2 and 5, `references/closure.md`, and `CONTEXT.md` (Clump reworded; Family added). `tier.py` and `runfile.py` grammar is unchanged.

Tests: a fixture where one hub file chains five candidates yields one family of five clumps; two candidates with identical closures yield one clump of two; four identical closures yield clumps of three and one; a dispatch test with one family member live shows the rest held and a clump from a different family dispatched; the JSON output round-trips through `read_clumps`.

Filed from conversation — the burn's controller, the run stuck on what this fixes, 2026-09-21.

## Blocked by

- None — can start immediately.

PR #1050, opened 2026-09-22
loop.py: nested herdr agent_status, landed-clump core charging, same-tick held lines

### LINKED ISSUE #971: loop.py dispatch: name the clumps the same-tick guard skips
`loop.py dispatch` prints no line for a clump that `picks`' same-tick guard skips. The guard is the `any(paths(clump) & paths(earlier) for earlier in picked)` check in `burndown/loop.py` `picks`. With nothing in flight, `frontier` marks every member of a family dispatchable. `picks` then silently skips each member that shares a file with a clump it already picked this tick. The output shows only the `dispatch` lines and idle slots, with no `held` line explaining why.

Evidence: the correctness review of #970 (finding C1) piped `closure.py --json` over a hub family and one independent ticket into `loop.py dispatch --free 4`. The output was `dispatch #1` and `dispatch #4`, two slots stayed idle, and #2 and #3 got no line. No two workers share a file, so the invariant holds. But `burndown/SKILL.md` § The loop step 5 says dispatch "names what holds the rest", and since #970 this path is the first tick of every family.

Want: `render_dispatch` names each clump the same-tick guard skipped, and the picked clump it collides with, in the shape of a `held` line. Seam: `picks` / `render_dispatch` and the CLI `dispatch` output. Tests go in `burndown/loop_test.py`.

Filed from: the #970 build's round-1 correctness review.

## Blocked by

- #970

### LINKED ISSUE #988: loop.py sweep reads agent_status at the top of herdr's result; herdr nests it under agent
`python3 burndown/loop.py sweep --workers <file>` reported every one of five live slots as `unknown ... herdr reports agent_status None` on run `burn-2026-09-21-0930` (agent-skills, 2026-09-21 14:58), while `herdr agent get <name>` answered for each of them. The probe reads `agent_status` off the top level of `result` (grep `agent_status` in `burndown/loop.py`); herdr 's `agent get` answer nests the agent's fields under `result.agent` — `python3 -c` over the same output finds `result["agent"]["agent_status"]` set to `working` / `idle` / `done`. So the sweep, the one backstop that can find a vanished pane (#778), reads every live pane as unmeasurable.

Fix: read the status from `result.agent` when present, top level otherwise, and keep `unknown` for a shape that has neither; add a test with herdr's real nested shape (paste one `herdr agent get` answer into the fixture) beside the flat one. The `loop.py` sweep tests today cover only the flat shape, which is why this passed.

Filed from conversation — the burn's controller, first sweep of the run, 2026-09-21.

## Blocked by

### LINKED ISSUE #1003: loop.py dispatch charges a landed clump's job before filtering landed clumps out
`burndown/loop.py` `dispatch` (grep `core_room`, from #933 / PR #1002) passes every `--in-flight` entry to the core accounting before it filters landed entries out of the live set. A clump the run file created with `job: null` and later landed (`land()` sets only `landed`) therefore refuses the whole tick with "live with no job record" instead of being read as a freed slot. The controller on run `burn-2026-09-21-0930` never hit it only because it hand-edited landed clumps out of the in-flight file before each tick; a controller that hands `dispatch` the run file's clumps as they stand would be refused after every landing. Raised by PR #1002's final Codex pass.

Fix: filter landed clumps out first and feed that one collection to both the core accounting and the peak live count; add a regression test with a run-file-style landed entry whose job is null, expecting the tick to proceed and count that slot as free.

Filed from PR #1002's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #971
Closes #988
Closes #1003

## What changed

- `loop.py sweep` reads herdr's `agent_status` from `result.agent` when nested there, falling back to the flat `result.agent_status` shape, instead of only the flat shape (#988).
- `loop.py dispatch` filters landed clumps out of `--in-flight` before charging core accounting or counting the peak live set, instead of charging a landed clump's null job record and refusing the tick (#1003).
- `loop.py dispatch`'s `picks` now returns the clumps its same-tick collision guard skipped, each tagged `same_tick` and naming the earlier pick it collided with; `render_dispatch` prints those as `held` lines instead of silently dropping them (#971).

## Tests run

`python3 burndown/loop_test.py` — 95 passed.

## Decisions made

Built as a dependency-ordered clump (971 depends on the now-landed #970; 988 and 1003 are independent) in one workspace/branch/PR per the dispatch instruction.

Round-1 findings from `/multi-axis-code-review` (standards, spec, correctness — 3 opus axes), each with its disposition:

- **P1** (spec, hard) / **C1** (correctness, hard) — same finding from two axes: `burndown/SKILL.md` § The loop step 5 still said the same-tick guard's skip "prints no held line," which #971 makes false. Fixed in `76bf462`.
- **S1** (standards, judgement) — `render_dispatch` duplicated its held-line format string behind a `"workspace" in held` discriminator; a same-tick entry that ever acquired a `workspace` key would misprint as a live-workspace hold. Fixed in `76bf462` — `picks` now tags same-tick entries `same_tick: True` explicitly, and `render_dispatch` computes one `where` clause.
- **S3** (standards, judgement) — `run()` synthesized a one-key `{"held": ...}` dict just to satisfy `render_dispatch` reading one key off a whole frontier state. Fixed in `76bf462` — `render_dispatch` now takes the held list directly.
- **S2** (standards, judgement) — filtering landed clumps out before `core_room` also means `core_room`'s only caller of `job_cores` never validates a landed clump's `job` record, not just never charges it. Disputed: a landed clump's `job` is stale by definition once the worker is gone (`land()` never clears it), and #1003's own repro is a landed clump with `job: null` that must not fail validation — exempting landed clumps from validation, not only from charging, is what the ticket asks for.
- **P2** (spec, judgement) — #988 asked to paste a real `herdr agent get` answer into the test fixture; the new nested-shape fixture is hand-reduced. Disputed: no live herdr agent was reachable to capture a literal answer in this build environment, and the pre-existing flat-shape fixture in the same file is likewise hand-built, so this matches the file's own precedent.
- **P3** (spec, judgement) — `picks` breaks its loop at the free-slot cutoff, so a same-tick collision among candidates past that cutoff is never examined or reported, narrower than SKILL.md's "names what holds the rest" at low `--free`. Filed as #1049.

Verification pass (one opus pass, scoped to the round-1 findings and fix commit `76bf462`) confirmed all four fixed claims against the current tree, re-ran the suite itself (94 / 77 green), passed S2/P2/P3 through as settled, and surfaced one new low-severity observation (V1: the live-workspace held-line wording had no full-string assertion after the S1 refactor) — closed in `63088af` by asserting the full line and mutation-checking it red.

**Codex adversarial-review gate, first pass (`fe15f40`).** One `[high]` finding, confirmed by the controller: `unlanded` was computed for `core_room` and the peak live count, but both dispatch-time `frontier()` calls still took the unfiltered `in_flight`, so a landed clump's dead workspace (its change is on `main`; the next worker branches from there) still blocked a candidate sharing its closure with a spurious `held` line. Fixed in `fe15f40` — both `frontier()` calls now take `unlanded`; new regression test asserts a candidate sharing a landed clump's closure dispatches with no `held` line.

## Last reviewed sha

`fe15f40` (Codex-gate fix). Nothing landed after it.

Gate pass on 63088af (needs-attention, one finding):
- [high] landed clumps still passed to both dispatch-time `frontier` calls, so a dead workspace held candidates: **fixed** fe15f40 — both calls take `unlanded`; regression test added.

Second pass on fe15f40: approve, no findings.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: landed clumps still participate in frontier collision checks, so a freed workspace can indefinitely block a candidate that shares its closure.

Findings:
- [high] Landed clumps still hold the dispatch frontier (burndown/loop.py:820-831)
  The new `unlanded` collection is used for core and peak accounting, but both frontier calls still receive `in_flight`. Thus a run-file-style clump with `landed` set and a closure overlapping a candidate produces a `held ... in /w/...` line and no dispatch, despite the code defining it as no longer live. This leaves a slot falsely blocked until some separate cleanup removes the record; the added #1003 regression misses it because its candidate uses `fresh.py` rather than the landed clump's `verify.py`.
  Recommendation: Pass `unlanded` to both `frontier` calls, and add a CLI regression where a landed, null-job clump and the candidate share a closure; assert the candidate is dispatched and no held line names the landed workspace.

Next steps:
- Fix the frontier inputs and run `python3 burndown/loop_test.py` in a writable test environment.

caneff:
# Codex Adversarial Review

Verdict: approve

Ship: no material defect found in the origin/main diff. The nested-status, landed-clump, and same-tick hold paths are correctly wired; targeted pure-path checks passed. Full gate could not run because this review environment makes /tmp read-only.

No material findings.

PR #1051, opened 2026-09-22
merge-cleanup: fail-closed reap re-check + clump sort witness (#975, #983)

### LINKED ISSUE #975: merge-cleanup: a failed worktree listing reads as no holder
`worktree_holding` in `flow/lane/src/bin/merge_cleanup.rs` is `quiet_stdout(...).unwrap_or_default()`, so a `git worktree list --porcelain` that fails returns the same empty string as a branch that no worktree holds. `reap_one`'s destructive-time scope re-check (#881) reads that empty string as "no holder" and goes on to remove the workspace and delete the branch — defect class 1 (an absent answer read as a benign one) at the call the re-check exists to guard.

Intent: tell "listed, no holder" from "could not list", and refuse on the second. A branch whose worktree simply vanished (listed, no holder) must still reap, as `a_branch_with_no_holder_at_all_since_the_plan_still_reaps` pins.

Pre-existing: the old `linked_worktree_holding` read the same empty string as `None`. Filed from round-1 review (S1, C1) of the #875/#881/#949 clump.

## Blocked by

- #881

### LINKED ISSUE #983: Witness merge-cleanup's length-then-text ticket ordering
`clear_landed_claims` in `flow/lane/src/bin/merge_cleanup.rs` sorts its string-keyed ticket set by length then text (grep `tickets.sort_by(|a, b| a.len().cmp(&b.len())`), and its comment claims that ordering is why the keys are digit strings. No test witnesses it: replacing it with `tickets.sort()` leaves the `merge_cleanup` suite green, because every clump fixture uses equal-width numbers. A clump fixture with mixed-width numbers (e.g. 9 and 10) asserting `cleared #9, #10` would fail on a lexicographic sort.

Parent: #889. Filed from #903's correctness review (finding C1).

## Blocked by

caneff:
Closes #975
Closes #983

## What changed

- `worktree_holding` now returns `Result<Option<String>, ()>` instead of an empty string standing in for both "listed, no holder" and "could not list"; `reap_one`'s destructive-time re-check refuses outright on a failed `git worktree list` instead of falling through to a messier downstream failure.
- Added a fixture that fails only the `git worktree list --porcelain` subcommand mid-run, proving the re-check now refuses cleanly instead of reaching a `FAILED` row via `git branch -d`'s own backstop.
- Added a mixed-width ticket-number fixture (`9`, `10`) witnessing `clear_landed_claims`'s length-then-text sort, which no prior clump fixture (all equal-width) actually exercised.

## Tests run

`cd flow/lane && cargo test` — 119 passed in `merge_cleanup`, plus 7 in `resolve_controller`, 0 failed.

## Decisions made

Round 1 (3 axes, 8 findings, all judgement/informational — no hard violations, no confirmed bugs):

- **S1** (standards) — `linked_worktree_holding`'s doc comment justified its fail-open reading only via `cleanup_branch`'s `git branch -d` backstop, but `worktree_plan` (the sweep row) shares the same fail-open path with no such backstop. Fixed: split the doc comment by call site, sha `4864d1b`.
- **S2** (standards) — the new listing-failure test is another hand-rolled shell-script write reaching into fixture internals. Disputed: pre-existing pattern across several other tests in this file; a repo-wide test-helper extraction is out of scope for this ticket.
- **S3** (standards) — the #983 witness test's name overclaimed "ascending numeric order" when the code only guarantees length-then-text (equal to numeric order for digit strings without leading zeros). Fixed: renamed to `a_clump_with_mixed_width_ticket_numbers_sorts_shorter_before_longer`, sha `4864d1b`.
- **P1** (spec) — #975's stated worst case ("removes the workspace and deletes the branch") doesn't match the landed evidence: `git branch -d`'s own check backstops this exact repro, so the actual pre-fix symptom is a `FAILED` row, not a deleted branch. Disputed: the fix is still correct and required regardless of which exact symptom the unpatched code hits; noted here rather than in code.
- **P2** (spec) — same underlying gap as S1 (the doc comment read as covering both `linked_worktree_holding` call sites when it only covered one). Fixed alongside S1, sha `4864d1b`.
- **C1** (correctness) — an old normalization quirk (`!wt.is_empty()`) is now `Ok(Some(""))` for the same malformed porcelain-line ordering. Disputed: unreachable with real git's actual output, and every consumer fails closed on it regardless.
- **C2** (correctness) — the sweep/`cleanup_branch` path still surfaces a failed listing as a `FAILED` row rather than a clean refusal, unlike `reap_one`'s new one. Disputed: #975's own documented, deliberate scope boundary — nothing is removed out of scope, only reported noisily. Possible follow-up ticket.
- **C3** (correctness) — the generated shell stub in the new #975 test interpolated paths unquoted. Fixed: quoted both interpolations in both generated scripts, sha `4864d1b`.

Verification pass confirmed every FIXED disposition against the actual diff hunks and every DISPUTED reasoning against the source; nothing was re-raised as still-live. Full dispositions: `~/.cache/agent-reviews/skills/dispositions-975-983.jsonl`.

## Last reviewed sha

`4864d1be3935559f4f0d29b0f48c670adf8ca3a1` — no commits past this sha.


caneff:
# Codex Adversarial Review

Verdict: approve

Ship: #975 now refuses the destructive-time reap when `git worktree list` fails, while preserving the successful no-holder reap path. #983’s mixed-width fixture witnesses the required length-then-text ordering.

No material findings.

Issue #934, opened 2026-09-20
Nothing enforces the commit-identity rule; the first failure is at GitHub's edge


caneff:
The rule landed today in `flow/claude/CLAUDE.md` under Hard rules: commit identity comes from the checkout, never from session context — never pass `-c user.email` / `-c user.name`. It is a sentence, and nothing in the repo enforces it.

Three workers broke it in one day (#887, #890, #893) by passing `-c user.email=<the user's real address>` on every commit, inferring a git identity from the harness context line that names the address for identification and attribution. Every git config layer was already correct: global `user.email t@t`, repo and worktree `t@example.com`, `git var GIT_AUTHOR_IDENT` resolving to `t <t@example.com>`. GitHub's email-privacy block rejected all three pushes with `push declined due to email privacy restrictions`, and each needed a history rewrite over unpushed commits — a gated operation — to undo. Root cause is recorded on #909.

The rule binds new sessions only and rests on a model reading it. The next model that reasons past it fails the same way, at the same place: GitHub's edge, after the work is committed, where the cheapest fix is the gated one.

What to build: a local check that refuses a commit whose author or committer email is not the configured `user.email` for that checkout, so the failure happens at commit time with the work still uncommitted and no rewrite needed. Decide and state where it fires — a `pre-commit` hook is per-clone and a worker's worktree is created by the lane, so the hook has to be installed by whatever creates the worktree (`flow/lane`), not assumed present. Say what happens when someone legitimately needs a different identity.

Anchors: `flow/lane/src/bin/implement_dispatch.rs` (worktree creation), `flow/claude/CLAUDE.md` Hard rules (the rule text), `implement/pre-report-gate.sh` (the existing pre-report check, as a model for a refusal that names itself).

Filed from conversation — root-cause investigation of the #893 push rejection, map #776.

## Blocked by

None — can start immediately.

PR #1053, opened 2026-09-22
implement-dispatch: bound the claim-lock fetch and stop trusting foreign pre-commit hooks by text

### LINKED ISSUE #976: Bound the git fetch that runs under implement-dispatch's claim lock
Parent: #905

`implement-dispatch` holds the per-repository claim lock (added in #905) from the first ticket read to the last claim edit, and that span includes `git fetch -q origin`, which runs through `runner::quiet_ok` with no timeout. A stalled fetch pins the repo's claim lock for the hung run's whole life, and every other dispatch of that repo refuses after the 30s lock wait. Fix by bounding the fetch with a timeout, moving it outside the lock, or stating the limit in the `--help` text.

Where: `flow/lane/src/bin/implement_dispatch.rs`, grep `"fetch", "-q", "origin"` and `ClaimLock`.

Filed from the #905 correctness review (finding C3).

## Blocked by

- #905

### LINKED ISSUE #1009: implement-dispatch accepts a foreign pre-commit hook on a text match, not on proof the guard runs
`flow/lane/src/bin/implement_dispatch.rs` (grep `invokes_guard`, from #934 / PR #1007) refuses dispatch on a foreign pre-commit hook that does not invoke the identity guard, but accepts one on any non-comment line containing `commit-identity-guard`. A hook that echoes the name, calls the guard on an unreachable branch, or appends `|| true` passes, keeps control of pre-commit, and a later `git -c user.email=<other> commit` succeeds — the rejected-push failure #934 exists to stop, behind a dispatch that reported the guard installed. The accepted-hook path also never checks the executable bit, and git ignores a hook without it. Raised by PR #1007's final Codex pass.

Fix: stop accepting foreign hooks by source text. Either refuse every non-owned hook, or take ownership by moving the foreign hook aside under a stable name and installing the lane's wrapper, which runs it and then requires the guard to succeed; verify the executable bit on whatever is left in place. Tests: end-to-end commits under a spoofed executable hook, an `|| true` hook and a non-executable hook, each refusing a foreign-email commit.

Filed from PR #1007's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
**Closes #976**
**Closes #1009**

## What changed

- `implement-dispatch` bounds the claim-lock `git fetch` with a timeout (`LANE_FETCH_TIMEOUT_MS`, 120s default) so a stalled fetch can no longer pin the repo's claim lock forever.
- `implement-dispatch` no longer trusts a foreign `pre-commit` hook by matching text: it always takes ownership, displacing whatever is there to a stable name, forcing it executable, and running it before unconditionally running the commit-identity guard. Ownership itself is byte-identity with the wrapper's exact current text.
- `flow/claude/OPERATIONS.md` updated to describe the takeover instead of the old refuse-on-text-miss behavior.

## Tests run

`cd flow/lane && cargo test` — 90 passed in the integration suite, 3 passed in the in-binary unit tests, all other lane suites unaffected. 0 failed.

## Decisions made

Round-1 findings (multi-axis-code-review), each with its disposition:

- S1 (judgement) — fixed, `4232c41`: `pre_commit_wrapper()` builds the wrapper text from the `PRE_COMMIT_FOREIGN` const via `format!`, one source for the name instead of a duplicated literal.
- S2 (judgement) — fixed, `4232c41`: same fix as C3 (collision refusal covers the silent-clobber case).
- S3 (judgement) — fixed, `4232c41`: a read error on `pre-commit` other than `NotFound` now refuses install instead of being read as "foreign".
- S4 (judgement) — disputed: only two env-duration-getter instances exist (`claim_lock_wait`, `fetch_timeout`) — rule of three not met for extracting a shared helper. The "orphaned sleep" concern is unfounded: `run_bounded`'s `wait_bounded` kills the whole process group on timeout. The test's sleep was shortened anyway as part of C6.
- P1 (hard) — fixed, `4232c41`: `flow/claude/OPERATIONS.md` rewritten to describe the takeover.
- P2 (judgement) — fixed, `4232c41`: `install_identity_guard` now prints a one-line notice naming both paths when it takes ownership.
- P3 (judgement) — fixed, `4232c41`: same fix as C3.
- P4 (judgement) — fixed, `4232c41`: `refuses_a_foreign_email_commit` now also asserts the guard's own message appears in stderr, not just a nonzero exit.
- C1 (hard) — fixed then revised: `4232c41` first fixed it with a marker-based ownership check ("does the file contain a stable comment substring"), which itself reopened #1009's hole — a foreign hook merely carrying that phrase in a comment would read as "ours" and skip the takeover entirely, so the guard would never install. Caught on the Codex adversarial-review gate for this PR; fixed for real in `b527001` by reverting to byte-identity ownership (the original #934 ruling) and accepting that a stale prior version of the wrapper is foreign too — displacing and running it is harmless, since its own body is the pre-#1009 two-line form with nothing self-referential in it.
- C2 (hard) — fixed, `4232c41`: `install_identity_guard` now runs immediately after `ClaimLock::acquire` succeeds, so two concurrent dispatches of one repo can't race each other's takeover.
- C3 (hard) — fixed, `4232c41`: a `pre-commit.foreign` already holding a *different* foreign hook refuses the dispatch by name instead of being silently overwritten. New regression test `a_pre_commit_foreign_already_holding_a_different_hook_refuses_dispatch`.
- C4 (judgement) — fixed, `4232c41`: the foreign hook is copied aside (original left in place) rather than moved, so `write_executable`'s own atomic rename onto `pre-commit` is what replaces it — no window with no hook at all.
- C5 (judgement) — disputed: forcing the displaced hook's executable bit is required, not optional. The wrapper body is `"$dir/pre-commit.foreign" "$@" || exit $?` before `exec ".../commit-identity-guard"` — an unforced, non-executable `pre-commit.foreign` fails to exec, and `|| exit $?` exits the wrapper before the guard ever runs, blocking even a correctly-configured commit. #1009's own acceptance criteria require the non-executable-hook scenario to both refuse a bad-email commit and allow a good one; forcing the bit is what makes that reachable. (Verification pass checked this by reverting the chmod in a throwaway worktree and confirmed the good-commit case then fails with a permission-denied exec error.)
- C6 (judgement) — fixed, `4232c41`: the stalled-fetch test's fake `git` sleeps 30s instead of 600s.
- C7 (judgement) — fixed, `4232c41`: same fix as P4 (shared test helper).
- Residual (found by the verification pass, not a round-1 id, judgement) — fixed, `fe8ed18`: the C3 collision check's read of `pre-commit.foreign` also silently swallowed non-`NotFound` errors; given the same `match` treatment as `pre-commit` itself.
- Codex gate finding (`[high]`, PR #1053, controller-confirmed) — fixed, `b527001`: see C1 above. Two new end-to-end tests added: `a_hook_that_spoofs_the_wrapper_comment_is_still_taken_over_and_the_guard_still_runs` and `a_non_executable_hook_that_spoofs_the_wrapper_comment_is_still_taken_over_and_the_guard_still_runs`, plus `a_stale_lane_wrapper_with_different_bytes_is_displaced_like_any_foreign_hook_and_the_guard_still_runs` covering the harmless-displacement case byte-identity relies on.

## Last reviewed sha

`4232c41` was the sha the multi-axis-code-review verification pass reviewed. `fe8ed18` and `b527001` landed after — `b527001` in response to the Codex gate's `[high]` finding — and are unreviewed by the Claude-lane axes past `4232c41`.

Gate pass on fe8ed18 (needs-attention, one finding):
- [high] `is_ours` trusted a marker string, so a foreign hook carrying it skipped takeover and the guard never ran: **fixed** b527001 — ownership is byte-identity with the current wrapper (the #934 ruling), exec bit verified after every write, two marker-spoof end-to-end tests.

Second pass on b527001 (needs-attention, one finding):
- [high] a byte-identical owned wrapper that lost its exec bit is not repaired on a later dispatch: **filed** as #1054. No third run: controller disposition.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the #1009 takeover still trusts a foreign hook by source text, allowing the identity guard to be bypassed.

Findings:
- [high] A foreign hook can spoof wrapper ownership and bypass the guard (flow/lane/src/bin/implement_dispatch.rs:397)
  `is_ours` treats any readable pre-commit containing the marker as the lane wrapper. A foreign hook can include that phrase in a comment and exit successfully (or otherwise omit the guard); dispatch then skips both takeover and wrapper installation. Git will execute that hook, so a worker can commit with a foreign email despite dispatch reporting success. This is the same text-based trust boundary #1009 requires eliminating; a non-executable spoof also bypasses the required executable-bit check.
  Recommendation: Do not classify arbitrary hook contents as lane-owned. Validate and repair an owned wrapper's complete, expected behavior and executable mode, or use ownership state that cannot be supplied by the foreign hook; otherwise treat it as foreign, preserve it, and install the wrapper. Add end-to-end tests for marker-spoofed executable and non-executable hooks that attempt a foreign-email commit.

Next steps:
- Fix ownership classification and rerun the identity-guard end-to-end tests.

caneff:
# Codex Adversarial Review

Verdict: needs-attention

No ship: an owned but non-executable wrapper bypasses the identity guard on later dispatches.

Findings:
- [high] Byte-identical owned wrapper is not checked or repaired for executability (flow/lane/src/bin/implement_dispatch.rs:402-432)
  After a successful dispatch, a permission change can leave `.git/hooks/pre-commit` at mode 0644 while preserving its exact bytes. A later dispatch sets `is_ours` true at line 402, refreshes only the guard, and skips `write_executable` for `pre-commit` at line 431. Git ignores a non-executable hook, so a subsequent `git -c user.email=<foreign> commit` succeeds despite dispatch reporting the guard installed. This violates #1009's executable-hook requirement; current tests cover a non-executable foreign hook, not a non-executable byte-identical lane wrapper.
  Recommendation: When the existing wrapper is byte-identical, stat and require an executable bit; preferably repair it through the same atomic `write_executable` path. Add an end-to-end test that chmods an installed lane wrapper non-executable, dispatches again, and proves a foreign-email commit is refused.

Next steps:
- Fix the existing-owned-wrapper executable-bit path and add the regression test.

PR #1055, opened 2026-09-22
check-section-references: unique heading resolution, correct inline-code guard

### LINKED ISSUE #990: check-section-references: a step reference can resolve in a sibling heading that shares the prefix
`tests/check-section-references.py` (grep `section_has_steps`, from #913 / PR #986) collects every heading whose text starts with the reference label and accepts the reference if any candidate section contains the requested step. So `§ Merge step 6` stays green after step 6 is removed from `# Merge` as long as `# Merge notes` holds a `6.` item — the drift #884 and #913 exist to catch, escaping through prefix matching. Raised by PR #986's final Codex pass; the fixture set has no collision case.

Fix: resolve a reference to one heading before checking steps — an exact normalized match wins; otherwise a prefix match must be unique, and an ambiguous one is refused with a message naming the candidates — and run `section_has_steps` only on that heading. Add an invalid fixture with two prefix-sharing headings where only the sibling holds the step, and a valid one where the exact heading holds it.

Filed from PR #986's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

### LINKED ISSUE #1005: check-section-references: the inline-code guard counts single backticks and runs after the trimmers
`tests/check-section-references.py` (grep `inside_span`, from #955 / PR #1004) refuses a reference whose name is cut at inline code, so an inline-code heading name cannot silently downgrade to a prefix match. Two gaps, raised by PR #1004's final Codex pass:

1. `inside_span` counts single backticks modulo two instead of matching the opening delimiter run, so a valid double-backtick span (``§ Build`` inside a two-backtick code span) reads as outside a span and is refused.
2. The guard compares the raw capture against the already-trimmed fallback label, so a name the possessive or conjunction trimmer shortens before the backtick (`§ Build's `release``) becomes `Build`, skips the guard, and can prefix-match an unrelated `## Build` heading — the silent downgrade the guard exists to stop.

Fix: parse inline-code delimiter runs and require a matching closing run; apply the guard to the name before any prose trimmer touches it; add fixtures for a double-backtick-wrapped reference (valid) and for possessive and conjunction text immediately before a backtick (refused). Re-run the repository sweep afterwards; #955's baseline was 90 references with every resolved name identical.

Filed from PR #1004's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #990
Closes #1005

## What changed

- `resolve_heading` now resolves a `§` reference to one heading before checking steps: an exact normalized match wins over any number of headings sharing the name's prefix, and a tie (no exact match, more than one prefix match, or more than one exact match) is refused, naming every candidate.
- The inline-code guard now parses real backtick delimiter runs (an opening run closes only on a later run of the same length, so a double-backtick span is read correctly) and applies to the name before the possessive/conjunction prose trimmer touches it, while staying exempt when a step locator already ends the name on its own terms.
- New fixtures cover both fixes and the paths round-1 review found untested: a sibling-prefix step collision, an exact match winning over a sibling, a genuine prefix ambiguity, a duplicated exact heading, a double-backtick-wrapped reference, possessive/conjunction text immediately before a backtick, and a step locator legitimately followed by a backtick.

## Tests run

`bash tests/all.sh` — all 77 suites pass (includes `tests/section-references.test.sh`, which runs the real repository-wide sweep first).

## Decisions made

Clump of two tickets (#990, #1005), built in dependency order in one workspace/branch/PR per the controller's brief; no per-ticket sha is pinned since neither survives the squash merge.

One unrequested-looking change: a word-wrap-only reflow of one sentence in `implement/SKILL.md` (the "PR up" controller trailer). Its original wrap split "§ The" from "merge" across a line break; the `§` checker reads per-line, so that reference read as the bare label "The," which the #990 fix (correctly) now flags as ambiguous against "The brief" / "The PR" / "The merge" in the same file. Rewrapped so "§ The merge" stays on one line; wording and the pinned substring `answer my outstanding questions` (required by `implement/pr-up-report-shape.test.sh`) are unchanged. Necessary to keep the branch's own gate green under the ticket's own fix, not scope creep.

Round-1 findings from `/multi-axis-code-review` (10 total: 2 hard, 8 judgement), each disposed and confirmed by a separate verification pass:

- **S1** (judgement) — disputed: the step-locator-exemption check is a list-length read on a value (`readings`) now bound once per pointer rather than recomputed (S2); making it a fully explicit boolean would change `reference_targets`' return contract across three call sites for marginal robustness on one boolean in a 300-line test script.
- **S2** (judgement) — fixed, `394e96c`: `reference_targets(pointer)` is called once per pointer and the result (`readings`) reused at all three use sites.
- **S3** (judgement) — disputed: the per-pointer body in `main()` is long with a substantial inline comment, but the comment records *why* #1005's bug happened; lifting the block to a helper relocates that explanation rather than removing complexity, for a block with one call site.
- **S4** (judgement) — disputed: `resolve_heading`'s `candidates`/`ambiguous` are read after a `for label, steps in readings:` loop without pre-initialization, but `reference_targets` unconditionally appends its plain fallback reading at the end of every code path, so the list is never empty and the loop always binds them — the theoretical `UnboundLocalError` is unreachable as written.
- **Over-engineering note** (no sidecar id) — disputed: `code_spans`' full delimiter-run matching (needed for the double-backtick case) was flagged as possibly yagni since only the new fixture exercises it; #1005 explicitly asks to "add fixtures for a double-backtick-wrapped reference (valid)," so this is spec-mandated, not speculative.
- **P1** (hard) — fixed, `394e96c`: the doc paragraph added for #1005 used literal `§` in its own examples, which the checker itself read as four unresolved `§ Build` references — the branch's own gate was red. Rewritten using the doc's existing `<sign>` placeholder convention.
- **P2** (judgement) — fixed, `394e96c`: `resolve_heading`'s exact-match branch returned every exact match's body without checking there was only one, so two identically-normalized headings in one file would read "any of them holds the step" — #990's own escape, reopened inside the branch written to close it. Now refused as ambiguous; new fixture `step-duplicate-exact-invalid`.
- **C1** (hard) — fixed, `394e96c`: same root cause and fix as P1, confirmed independently by the correctness axis with its own repro.
- **C2** (judgement) — fixed, `394e96c`: the new "is ambiguous" failure path had no dedicated fixture (only witnessed by one real line in `implement/SKILL.md`); added `step-prefix-ambiguous-invalid` (genuine prefix tie, no exact match) and `step-duplicate-exact-invalid`.
- **C3** (judgement) — fixed, `394e96c`: the step-locator exemption in the inline-code guard had no dedicated fixture; added `step-locator-inline-code-valid`.
- **C4** (judgement) — disputed: `tracked_markdown()` returning zero files (and the checker exiting 0) for an untracked fixture directory was flagged as a defect-class-1 risk, but real invocations always run against a committed repo, and every existing `run_fixture` call already stages its fixture with `git add -A` before invoking the checker — the vacuous-pass path can't occur through either real invocation path. Informational: fixture staging is a convention in the test harness, not an enforced guard, so a fixture added some other way could still pass vacuously — noted for awareness, not actionable in this diff's scope.

## Last reviewed sha

`394e96c` — the verification pass's own commit is included in what it verified (it checked out that sha and confirmed the fixes it describes). No commits landed after it.

Gate pass on 394e96c (needs-attention, one finding):
- [medium] a code span crossing a line boundary is not seen as code, so a `§` inside it is rejected: **filed** as #1056 — the per-line scan predates this PR (main already iterated lines); #1005's guard made the miss visible. Not this PR's regression; the fix round was not spent on it. Controller disposition.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No ship: valid multiline inline-code references are rejected, so the repository gate can turn main red on compliant Markdown.

Findings:
- [medium] Inline-code span detection loses spans that cross a line boundary (tests/check-section-references.py:59-89)
  `code_spans` only receives the current line, but CommonMark code spans may contain line endings. For a valid reference such as an opening delimiter on the preceding line followed by `§ Build` and its matching closing delimiter on the current line, this function finds no opener; line 306 therefore treats the reference as outside code, and lines 308-319 reject it as a name cut at inline code. The new documentation explicitly presents this as CommonMark delimiter handling, while the added fixture covers only a same-line double-backtick span. This creates a false-positive repository-wide gate failure for valid Markdown.
  Recommendation: Determine code-span membership against document-level text (or carry delimiter-run state across lines) before evaluating each reference, and add a valid fixture with matching single- and double-backtick delimiters spanning lines.

Next steps:
- Add the multiline-span fixture and make span detection document-aware; then rerun the section-reference suite and full sweep.

PR #1057, opened 2026-09-22
worker-stop-alert: fix Monitor-timeout liveness misread and herdr-name controller matching (#981, #1014)

### LINKED ISSUE #981: worker-stop-alert: a timed-out Monitor notification reads as liveness
`flow/claude/hooks/worker-stop-alert.sh` (grep `select(test("<status>") | not)` in the `$touched` block, landed by #900 / PR #979) counts every task-notification without a `<status>` tag as evidence that the launch it names is still out. A Monitor that times out emits exactly that shape: `<event>[Monitor timed out — re-arm if needed.]</event>` with no `<status>`. Real transcripts hold it (`grep -rl 'Monitor timed out' ~/.claude/projects --include='*.jsonl'` hits several sessions). So a Monitor that has stopped keeps its earlier-turn launch outstanding across a peer-started turn, the worker's next silent stop is classified `waiting`, and the controller gets no alert — the terminal-state-as-benign shape in `docs/agents/defect-classes.md` class 1.

Fix: read a statusless notification whose event text is the Monitor timeout as a finish, not a touch, before it reaches `$touched`; add a fixture in `worker-stop-alert.test.sh` with a timed-out Monitor notification followed by a silent stop that must alert, in the shape of the denied-poll and errored-poll fixtures #979 added.

Filed from PR #979's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

### LINKED ISSUE #1014: worker-stop-alert: resolve a herdr controller name in the brief before matching reports and finding the pane
`flow/claude/hooks/worker-stop-alert.sh` (grep `.name == $controller`) reads the brief's `--controller` value, looks it up as a Claude session name in the session records, and counts a report only when a SendMessage `to` equals that literal or a socket obtained from that lookup. After #923 (PR #1012) the brief carries the controller's herdr agent name, which a compliant worker resolves to a session name or socket before sending, so the two can differ: a delivered report can read as silent, and when an alert is owed the hook has no session or pane for the controller and exits after logging `not-sent`. On run `burn-2026-09-21-0930` the alert still fired and matched, because workers sent to the controller's socket and the hook accepts a socket from the lookup; that is the path the fix should make deliberate rather than incidental. Raised by PR #1012's final Codex pass.

Fix: resolve a herdr-name controller through `herdr agent list` and the live session registry before report matching and pane lookup, accept a report addressed to the resolved session name or its socket, and add a transcript fixture whose brief holds a herdr name and whose report targets the resolved session.

Filed from PR #1012's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #981
Closes #1014

## What changed

`flow/claude/hooks/worker-stop-alert.sh` had two independent bugs: a timed-out Monitor's statusless notification read as liveness evidence, keeping its launch outstanding across a peer-started turn (#981); and controller matching used the brief's `--controller` value directly as a session name, which no longer holds since the brief carries the controller's herdr agent name (#923) — a delivered report could read as silent, and the pane lookup could find nothing when an alert was owed (#1014). Both are fixed, with fixtures for each failure mode.

## Tests run

`bash flow/claude/hooks/worker-stop-alert.test.sh` — ALL PASS
`bash tests/all.sh` — 77 suites passed (run three times across the build)

## Decisions made

Round 1 findings (multi-axis-code-review), each with its disposition:

- **S1** (standards, judgement) — the new controller-resolution `herdr` call sat outside the hook's stated 12s-deadline/15s-timeout budget. Fixed, a54396e: reduced the resolution timeout to 2s and the post-verdict deadline to 10s.
- **S2** (standards, judgement) — a missing/failing `herdr agent list` degrades silently to the pre-#1014 name-only matching, with no log trace on the matching path (only the alert path logs). Disputed: the fallback is #1014's own stated fix text ("fall back to treating the brief's value as a session name directly"), and the alert path already logs `not-sent ... no herdr pane for controller <name>` when an alert is owed; extra matching-path logging for a merely-absent herdr is beyond this ticket's scope.
- **S3** (standards, judgement) — `resolve_session`'s three tab-joined fields depended on an ordering constraint enforced only by prose (an empty field must be last, or `read`'s tab-`IFS` collapses it into the next field); `stat`/`fields` also escaped `local`. Fixed, a54396e (local) and 4e6e765 (structural fix: joined/split on `\x1f`, which isn't IFS-whitespace, so no ordering constraint is needed at all).
- **S4** (standards, judgement) — Mysterious Name: a one-letter `c` held the resolved controller name beside `controller`, and the alert text still printed the briefed one. Fixed, a54396e: renamed to `resolved_name`.
- **P1** (spec, judgement) — the resolved session name replaced the brief's literal controller value entirely instead of being accepted alongside it, so a report addressed to the brief's own herdr name would misread as silent. Fixed, a54396e: matching now accepts either.
- **P2** (spec, judgement) — `resolve_session` requires a non-empty `.name`, so a herdr-resolved session whose live record has no name dropped the session id entirely and the pane lookup logged not-sent — exactly #1014's own problem statement. Fixed, 4e6e765: falls back to the herdr-resolved session id for the pane lookup even when nothing was found to match a report against.
- **Codex pass, PR #1057** (correctness, hard, confirmed) — the same non-empty-`.name` requirement in `resolve_session` also dropped the record's `messagingSocketPath`, so a report a worker delivered to a nameless controller's `uds:<socket>` had no socket to match against and read as silent, even though a nameless live session still sends and receives cross-session messages. Fixed, b5a1542: dropped the non-empty-name filter in `sid` mode (`name` mode is unaffected, since it matches on `.name` itself); the report-matching jq already only uses the resolved name when non-empty, so a nameless record now falls through to socket-only matching instead of being invisible.
- **C1** (correctness, hard, confirmed) — hollow witness: the "report addressed to the resolved session name counts as reported" fixture passed even with #1014's resolution reverted, because a `silent` verdict with no resolvable pane also produces no prompt. Fixed, a54396e: added `expect_reported`, which also checks for a `not-sent` log line.
- **C2** (correctness, judgement) — the #981 fix's two mechanisms (excluding a timeout id from `$touched`, adding it to `$finished`) were witnessed only in the cross-turn case, where the `$touched` exclusion alone is sufficient. Fixed, 4e6e765: added a same-turn launch+timeout fixture, where only `$finished` can end it.
- **C3** (correctness, hard, confirmed) — the Monitor-timeout match tested the whole statusless notification for the substring "Monitor timed out", so a live monitor whose own tailed output merely mentioned that phrase read as its own end, firing a spurious alert. Fixed, a54396e: matches the exact `<event>[Monitor timed out — re-arm if needed.]</event>` marker.
- **C4** (correctness, judgement) — same root cause as S3: `resolve_session` guarded `.name` but not `.sessionId`, so a record with a name and a socket but no sessionId shifted fields. Fixed, 4e6e765 (same structural fix as S3).

A verification pass (opus, adversarial) checked all ten dispositions against the fix commits; it contested S3, P2, C2 and C4 as originally disposed, all four were then fixed as described above and re-verified by the same pass (each mutation-witnessed: reverting the fix reddens exactly the fixture that claims to catch it).

## Last reviewed sha

b5a1542 — no commits past it.

Gate pass on 4e6e765 (needs-attention, one finding):
- [high] a PID-live controller record with no `.name` was dropped, so a report delivered to its socket read as silent: **fixed** b5a1542 — nameless records keep their socket; fixture with a socket-delivered report asserting `reported`.

Second pass on b5a1542 (needs-attention, one finding):
- [high] when two live records share a sessionId, the first nameless one wins over the later named one, so a report to the named session is missed: **filed** as #1059. Cost is a false stop alert, which is a hint the controller reads, not an action. No third run: controller disposition.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: #1014 still misses delivered reports sent to a resolved controller socket when its live registry record lacks a name.

Findings:
- [high] Nameless live controllers lose their socket report address (flow/claude/hooks/worker-stop-alert.sh:72-74)
  `resolve_session` filters out every live record without `.name` before returning its socket. The later `ctl_session=$herdr_sid` fallback restores only pane lookup, not `ctl_socket`; therefore a worker report to `uds:<messagingSocketPath>` is not included in `$sends` and is classified silent, prompting a controller that already received the report. This contradicts the ticket’s required session-name-or-socket matching and is not covered by the nameless-record fixture, which tests only the silent path.
  Recommendation: Resolve a PID-live record’s session id and socket independently of its name; retain its socket for report matching even when the name is empty. Add a fixture mapping a herdr controller name to a nameless live session record with `messagingSocketPath`, then send a successful report to `uds:<socket>` and assert a genuine `reported` verdict.

Next steps:
- Preserve and test socket-based matching for nameless live controller records.

caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: b5a1542 reintroduces a duplicate-session race that can classify a delivered report as silent and wake the controller unnecessarily.

Findings:
- [high] First nameless duplicate session masks the live named controller (flow/claude/hooks/worker-stop-alert.sh:78-87)
  `resolve_session sid` returns the first live record whose sessionId matches. The registry reader explicitly documents a race where two live records share a sessionId and deliberately scans for a later named record. After b5a1542, this hook accepts and immediately returns an earlier nameless record. A worker resolving the herdr name sends to the later record's session name; the hook has neither that name nor necessarily its socket, so it misses the successful SendMessage and emits a false silent-stop alert.
  Recommendation: For sid resolution, scan all live matching records and prefer a non-empty-name record, retaining a nameless socket record only as fallback. Add a fixture with two live records sharing the controller sessionId, the lexically first nameless and the later named, and assert a report to the named session is `reported`.

Next steps:
- Fix the duplicate-session selection and rerun the hook contract suite.

PR #1060, opened 2026-09-22
lane-fake: move the overlap-claim witness's barrier to the reread before the edit

### LINKED ISSUE #982: implement-dispatch: the overlap test's barrier sits at the first read, not the reread before the edit
`flow/lane/tests/implement_dispatch.rs` (grep `two_overlapping_dispatches_of_one_ticket_claim_it_once`, from #905 / PR #978) synchronizes the two dispatches at the fake gh's first `issue view`, then lets both run free. The claim's own critical section is the reread immediately before each edit. Without the lock, whether both runs reach their reread before either edits is a race on a fast path: the worker saw the no-lock mutation red 3 of 3, and PR #978's second Codex pass argued the same mutation can pass when one run completes reread-and-edit before the other's reread. Both are right — the witness is probabilistic, not deterministic — and `LANE_CLAIM_DELAY_MS`, the knob that used to widen the window, was removed in 5f24710.

Make the witness deterministic: move or add the barrier so both runs are held at the reread before the first claim edit, then confirm that removing the ClaimLock acquisition produces two claim edits every time, and that with the lock the second run refuses on `in-progress` every time.

Filed from PR #978's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #982

## What changed

- `view_barrier` in the test-only `gh` fake now synchronizes each dispatch at its *second* view of a ticket (the reread immediately before the claim edit — the actual critical section) instead of its first (the earlier refusal-pass read).
- The wait for the second dispatch's arrival is a bounded busy spin (~50ms) falling back to a 200µs sleep-poll, so the two releases land microseconds apart instead of up to 20ms apart, and the ordinary (lock-held) passing run doesn't peg a core for the full 3s deadline.
- Barrier I/O now fails loud (panics) instead of silently defaulting to "first view" on a dropped write or unreadable file, and the release marker is scoped per ticket number, not just per process.

## Tests run

`bash tests/all.sh` — 77 suites passed, including `flow/lane/Cargo.toml`.

Determinism verified by hand (not committed — a local, uncommitted flag on `ClaimLock::acquire` used only to loop the test under both conditions): the real (lock-held) test 15/15 green; the no-lock mutation 20/20 red — up from the worker's original 3/3 red the ticket describes as probabilistic. The verification pass independently re-ran this in an isolated worktree: 8/8 green, 8/8 red.

## Decisions made

Round 1 (3 parallel axes on 6d35015) found 9 findings plus 2 over-engineering notes. All disposed before the verification pass; sidecar at `~/.cache/agent-reviews/skills/dispositions-982.jsonl`.

- **S1 / P3 / C2** (defect class 1: barrier I/O silently fails open) — fixed, e650328843a514192e7efb2de516fd0c92198b6d
- **S3 / P2 / C3** (passing run hot-spins the full 3s deadline) — fixed, e650328843a514192e7efb2de516fd0c92198b6d
- **C1** (release marker counted across every ticket in the dir, not just the one being read) — fixed, e650328843a514192e7efb2de516fd0c92198b6d
- **P1** (doc comment overstated simultaneity) — fixed, 3e37299daf1e0d177d416f8a3254381dcf51176c
- **S-over1** (dead `mine.exists()` guard, unreachable given the monotonic counter) — fixed, e650328843a514192e7efb2de516fd0c92198b6d
- **S-over2** (shrink: put reread markers in their own subdir) — disputed: the `.{n}.reread` suffix filter was needed anyway to fix C1, and a subdir would be a second mechanism for the same scoping
- **S2** (magic literal `2` encodes production's exactly-two-reads assumption with nothing asserting it) — disputed in part by the worker (the passing/failing test outcomes would show a relocated barrier), contested in part by the verification pass (a relocated barrier degrades silently back to the pre-#982 probabilistic witness — the test's own assertions all still pass, since the claim lock alone produces them). Left open as a judgement call, not fixed: the manual mutation loop is the only thing that currently catches it, and it isn't in CI.

- **Codex gate, round 1 (finding 1, high)** — `view_barrier` ran before the row/claimed check was computed, so release only synchronized *arrival* at the reread, not the *decision*; each process still independently computed and returned its row post-release, leaving a real gap. Fixed, 6808910f4aed03052e12d888f5f3e2372644f996: compute the row (including the claim check) before calling `view_barrier`, so the barrier only holds the return of an already-decided answer.
- **Codex gate, round 1 (finding 2, medium)** — the reread-marker `count()` mapped a failed `read_dir` or a failed entry read to 0 and kept waiting, the same fail-open shape S1/P3/C2 closed, just as a timeout instead of a race. Fixed, 6808910f4aed03052e12d888f5f3e2372644f996: panic on both; only the 3s deadline is a legitimate reason to stop.

## Last reviewed sha

6808910f4aed03052e12d888f5f3e2372644f996 — no commits past this sha.

Gate pass on 3e37299 (needs-attention, two findings), both **fixed** in 6808910 by controller ruling:
- [high] the barrier released before the state read, so one run could read, edit and finish before the other read: the row and claim state are now captured before the barrier and returned after release.
- [medium] a failed read_dir in the marker count mapped to zero and timed out silently: it now panics, per the file's own fail-loud policy.

Second pass on 6808910: approve, no findings.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the revised fake still does not deterministically make both claim rereads observe the unclaimed ticket, so removing the lock can still pass this witness.

Findings:
- [high] Barrier releases before either fake performs the reread (flow/lane/src/bin/lane_fake.rs:114-177)
  `view_barrier` runs before `run_gh` reads the issue row and checks `GH_CLAIM_DIR`. Once two markers exist, either process may be scheduled through the post-barrier state read and its claim edit while the other remains descheduled. The latter then observes `in-progress`, refuses, and the existing one-edit assertion passes even if `ClaimLock::acquire` is removed. Thus the test remains a probabilistic lock witness, contrary to the required mutation behavior.
  Recommendation: Capture the issue state used for the second view before entering the two-party barrier, then release both fake-gh invocations to return that captured unclaimed state. Add or run a lock-removal mutation assertion repeatedly and require exactly two claim edits; retain the normal-lock assertion that exactly one run refuses on `in-progress`.
- [medium] Directory-scan failures silently disable synchronization (flow/lane/src/bin/lane_fake.rs:154-163)
  The marker count converts any `read_dir` failure to zero and drops individual entry errors. After the three-second timeout, the fake proceeds without proving that two rereads met. An asymmetric or transient scan failure can therefore let one no-lock dispatch claim before the other performs its state read, again yielding the false one-edit pass that this test is meant to exclude. This also contradicts the stated fail-loud barrier-I/O policy.
  Recommendation: Propagate `read_dir` and directory-entry errors as a test failure rather than mapping them to zero; distinguish a deliberate two-party timeout from an inability to inspect barrier state.

caneff:
# Codex Adversarial Review

Verdict: approve

Ship: the captured pre-barrier row plus per-dispatch, per-ticket second-view markers makes the no-lock path deterministically return two unclaimed rereads, while the lock path serializes the second dispatch until it observes in-progress. No material defect found.

No material findings.

PR #1061, opened 2026-09-22
worker-spin-alert: key the dedupe on the streak's run boundary

### LINKED ISSUE #998: worker-spin-alert: a second spin of the same call later in the session is deduplicated away
`flow/claude/hooks/worker-spin-alert.sh` (from #925 / PR #992) keys its once-per-run dedupe on session, tool and a digest of the input. A worker that spins on one call, is interrupted by a different tool call, and later spins on the same call again reaches N a second time but matches the first run's sent record and exits without alerting, so a resumed spin stays invisible for the rest of the session. Raised by PR #992's final Codex pass.

Fix: put a run boundary into the dedupe identity — the tool-use id of the first call in the current consecutive streak is the natural one — so each streak alerts once and a later streak of the same call alerts again. Add a transcript fixture for spin, different call, same spin, expecting two alerts, beside the existing once-per-run fixture.

Filed from PR #992's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #998

## What changed

`worker-spin-alert.sh`'s once-per-run dedupe now keys on `run_id`, the
tool-use id of the current streak's first call, so a spin that is
interrupted by a different tool call and later resumes is a new streak
and alerts again instead of matching the first streak's `sent` log line.
`run_id` degrades to `null` — the pre-#998 (session, tool, digest) key —
whenever the classifier's read window (line cap or byte cap) or a
transcript missing a tool-use `id` makes the streak's true start
unknowable, so a long or oversized spin still alerts once rather than on
every call.

## Tests run

`bash flow/claude/hooks/worker-spin-alert.test.sh` — ALL PASS (21 checks).

## Decisions made

Round 1 (`multi-axis-code-review`, head `c4433322de4a612ddccdfe438c09600bfb05e4e8`):

- **S1** — a missing tool_use `id` silently restored the pre-#998 key with
  no signal anywhere. `fixed`, `1bdf175466644b872d9e88c4972529ae16d1e0fe`
  (the degrade is now documented as deliberate and witnessed directly by
  `no-id-spin.jsonl`).
- **S2** — `--classify` header comment's output contract omitted `run_id`.
  `fixed`, `2f298ac4526d450ec296e326bf4ce201161a616e`.
- **S3** — `resumed-spin-1.jsonl` was a byte-exact prefix duplicate of
  `resumed-spin-2.jsonl`. `fixed`, `2f298ac4526d450ec296e326bf4ce201161a616e`
  (deleted; the test reuses the already-committed `real-spin.jsonl`, same
  digest under a different `run_id`).
- **S4** — unreachable `$l != null` guard inside the classify reduce.
  `fixed`, `2f298ac4526d450ec296e326bf4ce201161a616e`.
- **P1 / C1** (found independently by both axes, hard) — `run_id` was the
  earliest call *visible in the read window*, so a streak longer than the
  window slid the id every call and re-alerted on every call instead of
  once. `fixed`, `1bdf175466644b872d9e88c4972529ae16d1e0fe`. First pass at
  this fix (`2f298ac`) only guarded the 500-line window cap; the
  verification pass (below) found the same bug still reachable through the
  4 MB byte cap (a spin on large inputs can exhaust it in well under 500
  calls) and disputed the disposition. `1bdf175` covers both caps and adds
  a byte-cap-only witness (`byte-cap-40/41.jsonl`, via a new
  `SPIN_BYTE_CAP` test override) alongside the line-cap witness
  (`window-501/502.jsonl`).
- **P2 / C2** — no fixture witnessed a growing streak under the new key,
  so nothing could have caught P1/C1. `fixed`,
  `2f298ac4526d450ec296e326bf4ce201161a616e` (`window-501/502.jsonl`,
  extended by `1bdf175` with the byte-cap pair).
- **P3** — same stale `--classify` doc as S2. `fixed`,
  `2f298ac4526d450ec296e326bf4ce201161a616e`.

Verification pass (scoped to round 1's findings and the two fix commits,
`dispositions-998.jsonl`): confirmed S2, S3, S4, P2, C2, P3 fixed;
disputed P1, C1 and S1 as noted above. Those three were then fixed for
real in `1bdf175466644b872d9e88c4972529ae16d1e0fe`, mutation-checked
against the new byte-cap and no-id fixtures.

## Last reviewed sha

`2f298ac4526d450ec296e326bf4ce201161a616e` — the verification pass's
disputed findings (P1, C1, S1) were fixed in the next commit,
`1bdf175466644b872d9e88c4972529ae16d1e0fe`, which was not independently
re-reviewed. I confirmed it myself: mutation-checked both new conditions
(byte-cap OR-clause, no-id fixture) red-without/green-with the fix, and
ran the full repo suite (`bash tests/all.sh`, 77 suites) clean.

Gate pass on 1bdf175 (needs-attention, one finding):
- [high] a streak that filled the window or byte cap flipped its dedupe key to a null run_id and alerted a second time: **fixed** af58196 — a null run_id dedupes against any prior sent line for the same (session, tool, digest); the accepted degrade is stated in the header; before/after-crossing tests per cap.

Second pass on af58196 (needs-attention, one finding):
- [high] 500+ sidechain tool_use lines can push an alerted spin's first id out of the raw window without nulling run_id, re-alerting once: **filed** as #1064. No third run: controller disposition.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: a continuous spin emits a second alert when it first crosses either read cap, violating once-per-run deduplication.

Findings:
- [high] Crossing a read cap changes the dedupe key mid-streak (flow/claude/hooks/worker-spin-alert.sh:73-109)
  A normal streak first alerts with its real first tool-use id. Once the same uninterrupted streak exceeds the 500-call window or byte cap, classify changes `run_id` to null; the dedupe key therefore changes from `(session, tool, digest, first-id)` to `(session, tool, digest, "")`, which has no matching sent record and sends a second alert. The new tests only invoke the hook after the cap has already been crossed, so they prove the null-key plateau dedupes with itself but do not witness the transition from the earlier real-id key. This creates duplicate controller alerts for any long live spin.
  Recommendation: Retain or resolve the already-sent run identity when entering the unknown-boundary plateau, so a cap-induced null `run_id` aliases the alert sent earlier in the same streak. Add hook-level tests that fire once before crossing each cap and again immediately after crossing it, asserting one total alert.

Next steps:
- Fix the cap-transition dedupe identity and add transition coverage for both line and byte caps.

caneff:
# Codex Adversarial Review

Verdict: needs-attention

No ship: the new run ID can still become window-relative and re-alert one uninterrupted spin.

Findings:
- [high] Sidechain tool-use lines can hide the run boundary without nulling run_id (flow/claude/hooks/worker-spin-alert.sh:64-73)
  The 500-line cap is applied before jq discards sidechain entries, but the null decision tests only parsed non-sidechain calls. After a spin has already alerted, 500+ sidechain tool-use records can push its true first call out of `tail -n 500`; after 20 subsequent identical main calls, `$calls` is below 500 and `byte_capped` is false, so the hook emits a new visible call ID as a real run_id. The exact-key dedupe then misses the prior sent alert and prompts the controller again for the same uninterrupted spin. The existing padding test has only 100 sidechain calls, so it does not witness this boundary loss. Track whether the raw candidate-line window was saturated (before filtering) and return null run_id whenever it was, with a regression fixture that alerts before and after a sidechain flood.
  Recommendation: Propagate a raw-window-truncated flag into jq and include it in the null-run_id condition; add a hook-level test for an alerted spin whose start is displaced by more than 500 ignored sidechain tool-use lines.

Next steps:
- Fix the raw-window saturation detection and add the regression fixture before shipping.

Issue #1062, opened 2026-09-22
Same-indent-rule readers disagree: fence bound vs. heading bound


caneff:
`burndown/frontier.py`'s fence markers (`_FENCE`/`_FENCE_OPEN`) now open/close only at CommonMark's <=3-space indent (#999), but `burndown/closure.py`'s `_ANY_HEADING`/`_CLOSURE_HEADING` and `implement-spec/closing_ticket.py`'s `_ANY_HEADING`/`_SEAM_HEADING`/`_KEY` still bound with `^[ \t]*#{1,6}` — any amount of indentation counts as a real heading in those two readers, unlike frontier.py's own `_HEADING`/`_ANY_HEADING`, which are already paired with `visible()`'s `_QUOTED` filter. After a repo applies two different indent rules to the same Markdown document depending on which reader parses it: an indented heading (4+ spaces or a tab) that frontier.py would drop as quotation can still register as a real section heading in closure.py or closing_ticket.py.

Separately, the same `{0,3}` CommonMark fence-indent literal is now duplicated across `burndown/frontier.py` (several sites) and `tests/check-section-references.py:53`, each with its own prose restating the rule, rather than one shared fragment.

Fix: give closure.py's and closing_ticket.py's heading regexes the same indent bound frontier.py's `_HEADING`/`_ANY_HEADING` use (or route them through a shared helper), and consider factoring the `{0,3}` fence literal into one place all readers/tests import.

Filed from caneff/agent-skills#999's round-1 multi-axis review (Standards axis, finding S2), 2026-09-22.

## Blocked by

- None — can start immediately.

caneff:
Closed by Chris's ruling, 2026-09-22: a Codex finding is filed only when its failure can occur in our environment (this box, these repos, bodies people actually write). This one cannot, so it is recorded in its PR's Codex dispositions and goes no further.

Issue #922, opened 2026-09-20
The frontier reader takes the first Blocked by section, so quoted evidence wins


caneff:
## Parent

#885

## TL;DR

If a ticket quotes another ticket's blocker list as evidence, the tool that decides what is ready to work on believes the quote instead of the ticket's own answer.

## What to build

`blocked_by_section` returns the **first** visible `## Blocked by` heading's contents. A ticket whose body quotes another ticket — without fencing it — therefore has its real declaration silently overridden by the quotation.

A body with more than one visible declaration is **ambiguous**, and ambiguous is `unresolved`, not "whichever came first". That is the same posture the reader already takes toward silence: an answer it cannot determine is one it refuses to guess at.

Make the reader detect a second visible declaration and report `unresolved` naming the ambiguity, rather than picking one.

## Evidence

Reproduced against `burndown/frontier.py` as merged at `64f01fe`:

```
body = evidence quoting "## Blocked by\n\n- #906", then the real "## Blocked by\n\nNone — can start immediately"

blocked_by_section -> '- #906\n\nAnd here is the real declaration appended by the skill:'
section_blockers   -> ([906], None)
```

Two failures in one: the quotation wins over the ticket's own answer, **and** the section runs on past the quotation to swallow the prose beneath it, because a section ends only at the next heading. The ticket reads as blocked by a stale reference it never claimed.

This is not hypothetical. `/file-ticket` appends its declaration at the end of the body (#911), so any finding whose evidence quotes a ticket body — which is the normal shape of a finding filed during a build — supersedes it. Codex raised it against #911; the emitter can fence its own evidence, but every hand-written and historical ticket remains exposed, and the reader is the only place that covers all of them.

## Acceptance criteria

- [ ] A body with two or more visible `## Blocked by` declarations classifies as `unresolved`
- [ ] The reported reason names the ambiguity, distinct from "states nothing at all"
- [ ] A fenced or indented quotation of a declaration is still not visible and does not trigger it
- [ ] A body with exactly one declaration is unchanged
- [ ] The inline `Blocked by:` forms count toward the same rule
- [ ] Each case has a test that fails before the change

## Blocked by

None — can start immediately.

## Seams under test

- `blocked_by_section`'s return contract, against fixture bodies: one declaration, two visible declarations, one visible plus one fenced, and an inline form beside a section.

Filed from a Codex finding on #911, reproduced by controller `skills-dc` against the merged reader, 2026-09-20.

PR #1063, opened 2026-09-22
frontier.py: recognize a fence marker only at CommonMark's ≤3-space indent

### LINKED ISSUE #999: frontier.py: an indented fence marker inside a quotation opens a fence and swallows the real declaration
`burndown/frontier.py` (grep `def visible`, from #922 / PR #996) runs every line through `unfenced()` before dropping four-space-indented quotation lines. CommonMark reads an indented ``` as indented code, not a fence, but `unfenced()` opens a fence on it; when that quoted block has no closing marker, every later line is discarded, including a real unindented `## Blocked by` / `None`, and the ticket reads as unresolved instead of unblocked. Fail-closed (held off the frontier, never wrongly dispatched), so narrow, but it is the quotation-does-not-affect-the-declaration guarantee #922 states. Raised by PR #996's final Codex pass.

Fix: apply the indented-code filter before fence tracking, or recognise a fence only at CommonMark's at-most-three-space indentation, in the one `visible` test both readers share; add a fixture with an indented literal ``` inside a quotation followed by one real declaration, expecting unblocked. `unfenced()` is imported by closure.py and closing_ticket.py, so keep its own behaviour or change all three readers together.

Filed from PR #996's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #999

## What changed

- `burndown/frontier.py`: `_FENCE`/`_FENCE_OPEN` now recognize a fence marker only at CommonMark's <=3-space indent bound, so a 4-space-indented (or tab-indented) `` ``` `` inside a quotation no longer opens a fence that swallows every later line, real declaration included.
- `burndown/closure.py` and `implement-spec/closing_ticket.py` (both import `unfenced()` from `frontier.py`) switch their declaration readers from raw `unfenced()` to `frontier.visible()`, which pairs the narrower fence bound with `_QUOTED`'s indented-quotation filter — closing a gap the narrower bound would otherwise have opened for those two readers (round-1 finding, see below).
- `burndown/references/frontier.md` and `burndown/references/closure.md` now state the <=3-space fence-indent bound explicitly.

## Tests run

- `python3 burndown/frontier_test.py` — 74 passed
- `python3 burndown/closure_test.py` — 52 passed
- `python3 implement-spec/closing_ticket_test.py` — 27 passed
- `bash tests/all.sh` — 77 suites passed

## Decisions made

- S1 (judgement, `burndown/frontier.py`): the fix's own comment claimed `visible()`'s `_QUOTED` bound is a backstop, but `closure.py`/`closing_ticket.py` called raw `unfenced()` with no such filter, so the claim was false for them. **Fixed** in `dd41dca` — switched both to `frontier.visible()`.
- S1's over-engineering note (shrink the frontier.py comment since the cross-reference was inaccurate): **disputed** — dd41dca made the claim true for all three readers, so nothing inaccurate is left to shrink.
- S2 (judgement): `closure.py`'s and `closing_ticket.py`'s own heading regexes still accept any indent (unlike frontier.py's, paired with `_QUOTED`), and the `{0,3}` fence literal is duplicated between `frontier.py` and `tests/check-section-references.py`. **Filed** as caneff/agent-skills#1062 — out of scope for #999's fence-only fix.
- S3 (judgement): the original new test covered only a 4-space-indented opener, not tab. **Fixed** in `d085100` — added a tab-indented-opener test.
- P1 (judgement, spec): the ticket's own alternative ("keep its own behaviour or change all three readers together") wasn't actually satisfied for `closure.py`/`closing_ticket.py`. **Fixed** in `dd41dca`.
- P2 (judgement, spec): `frontier.md`/`closure.md` didn't state the <=3-space bound. **Fixed** in `d085100`.
- C1 (confirmed behaviour change, plausible impact): narrowing the fence bound let a 4-5-space-indented quoted example become a phantom declaration in `closure.py`/`closing_ticket.py` (demonstrated). **Fixed** in `dd41dca`.
- C2 (hollow coverage): no test covered an indented *closer* ending an unindented fence early. **Fixed** in `d085100` — added `test_an_indented_closer_does_not_end_an_unindented_fence_early`.
- C3 (confirmed, environment, defect class 1): a stale gitignored `__pycache__/frontier.cpython-312.pyc` in the shared worktree caused a false test failure during round-1 review. **Disputed** as a diff defect — not fixable from the diff; cache cleared.

All eight round-1 findings were re-verified in a separate verification pass against the fix commits; all dispositions above were confirmed, with C1/C2 additionally confirmed via independent mutation checks. No new findings surfaced in verification.

- Codex adversarial-review pass ([medium]): `closure.scanned_lines`' Markdown branch kept raw `unfenced()`, with the stated reason "a repo-wide scan wants every non-fenced line, indented or not" — which doesn't hold for Markdown, where an indented block is a code block by the same CommonMark rule `declaration_section` already applies. A doc showing `    #include ../x` as an indented example created a real edge and a spurious family. **Fixed** in `6075b14` — switched `scanned_lines`' Markdown branch to `visible()`, same as `declaration_section`; source files unaffected.

## Last reviewed sha

6075b14 — no commits past it.

- Gate pass (head d085100), 1 medium: `scanned_lines` read an indented `#include` example in a `.md` as a live edge. **Fixed** in 6075b14 (Markdown branch goes through `frontier.visible()`, fixture added, mutation-checked).
- Second pass (head 6075b14), 1 medium: `_QUOTED` misses a space-then-tab indent, so a quoted ` \tNone` under `## Blocked by` reads as unblocked. **Filed** as #1065, blocked by #999: pre-existing on main from #922, outside this PR's diff, and no third round.


caneff:
# Codex Adversarial Review

Verdict: needs-attention

No-ship: the changed fence bound exposes indented Markdown code samples to closure scanning as real includes.

Findings:
- [medium] Indented Markdown examples can create false include edges (burndown/closure.py:236)
  `scanned_lines()` still consumes raw `unfenced()` output. After `unfenced()` stopped treating a four-space-indented ``` as a fence opener, a Markdown snippet such as `    #include ../_shared/line-kind.js` is returned and matched by `include_edges()`. That makes a quoted example join its document to a real target's closure, producing spurious collision families/clumps and withholding otherwise independent work. The existing fenced-example test does not cover an indented code block.
  Recommendation: Filter indented Markdown code/blockquote lines before matching include directives (for example, use `visible()` for Markdown scan input if that matches the intended grammar), and add a closure fixture proving an indented `#include` example creates no edge.

Next steps:
- Run the closure and full repository suites after adding the indented-example regression test.

caneff:
## Codex adversarial pass (second, head 6075b14)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: mixed space-plus-tab indented code can still be read as a live `None` declaration and wrongly dispatch a ticket.

Findings:
- [medium] Mixed tab indentation bypasses the quotation filter (burndown/frontier.py:120-128)
  `_QUOTED` recognizes only four literal spaces or a tab in column zero. CommonMark expands tabs to tab stops, so `" \tNone"`, `"  \tNone"`, and `"   \tNone"` are four-column indented code lines, but all survive `visible()`. Under a real `## Blocked by` heading, `.strip()` then turns that quoted line into `None`; `classify()` places the ticket in `unblocked`. This reverses the intended fail-closed behavior. A direct reproduction classified `## Blocked by\n\n \tNone` as unblocked.
  Recommendation: Make indentation recognition column-aware (for example, treat zero to three spaces followed by a tab as indented code, or expand leading tabs before applying the fence/quotation bounds) and add fixtures for each mixed space-plus-tab prefix.

Next steps:
- Fix the mixed-tab indentation case and rerun the frontier, closure, and closing-ticket suites.

PR #1066, opened 2026-09-22
Share brief, registry and herdr code between worker alert hooks

### LINKED ISSUE #991: Share brief, registry and herdr code between worker alert hooks

caneff:
Closes #991

## What changed

- Extracted the four pieces `worker-spin-alert.sh` copied from `worker-stop-alert.sh` — the `/implement` brief-capture regex, the `~/.claude/sessions/*.json` registry loop with its `/proc/<pid>/stat` field-22 liveness check, `logline()`, and the herdr agent-name/pane lookup — into a new shared `flow/claude/hooks/worker-alert-lib.sh`, sourced by both hooks. Each hook keeps its own alert text, dedupe key, and hook-specific logic (`worker-stop-alert.sh`'s `entries()`/verdict jq and blocked-retry loop; `worker-spin-alert.sh`'s `--classify` mode).
- Fixed a deploy gap round 1 found: `flow/install.sh` symlinks the two hooks into `~/.claude/hooks/` but never installed the new lib, so the installed hooks' `BASH_SOURCE`-relative `source` resolved to nothing — both test suites only ever exercised the hooks from the repo dir. `install.sh` now links `worker-alert-lib.sh` alongside its two hooks, and `install.test.sh` asserts it.
- Both hooks fail loud instead of silently: they check the `source`'s exit status, and — after a Codex gate pass on the first PR head found a lib that sources cleanly but is missing a called function passes that check anyway — also check each `worker_alert_*` function they call with `declare -F`. Either failure logs a `lib-missing` line naming what's wrong (via a `lib_missing()` helper that needs none of the lib's own functions) and exits 0, instead of continuing with command-not-found helpers and a silent, permanent "no controller" exit — the exact hazard this ticket exists to close, one layer up. Both `*.test.sh` suites assert both paths directly.

## Tests run

`bash flow/claude/hooks/worker-stop-alert.test.sh && bash flow/claude/hooks/worker-spin-alert.test.sh && bash flow/install.test.sh` — all three print `ALL PASS`, exit 0.

## Decisions made

- S1/P1/C1 (hard, `flow/install.sh` never linked `worker-alert-lib.sh`) — fixed, sha `2e1bf5494ef5877e67216d1c28261ac450140e0f`.
- S2/C2 (hard/judgement, unguarded `source` lets a failed load exit 0 silently) — fixed, sha `2e1bf5494ef5877e67216d1c28261ac450140e0f`; regression test added at sha `c8a7c5222b10935eb1c738a8c55e5399d4cd5a0c`.
- S3 (judgement, duplicated `\x1f`/nameless-session/Codex-#1057 rationale in hook and lib) — fixed, sha `2e1bf5494ef5877e67216d1c28261ac450140e0f` (trimmed to a pointer).
- S4 (judgement, the `sid|name` flag argument promoted into the shared API) — disputed: pre-existing shape, `resolve_session` already took this argument before the extraction; not introduced by this diff.
- P2 (judgement, spin hook keeps name-only controller resolution while stop hook tries herdr-sid first) — disputed: real divergence but pre-existing (the spin hook never had a herdr-sid path), and unifying the two strategies is a behavior change the ticket didn't ask for.
- C3 (judgement, both suites only ever invoked the hooks from the repo dir, so the installed-layout precondition went untested) — disputed as originally scoped (a full second-HOME integration suite is materially larger than this ticket); the residual the verification pass flagged (the fail-loud path itself was untested) was fixed instead, sha `c8a7c5222b10935eb1c738a8c55e5399d4cd5a0c`.
- Codex gate pass on `c8a7c522` (medium, confirmed): the fail-loud guard only caught a missing lib file, not a lib that sources cleanly but is missing a function the hook calls — fixed, sha `c7511eaba81d190ec30717708e09e1ecf8f4712b`, with a test case per hook.

## Last reviewed sha

`c7511eaba81d190ec30717708e09e1ecf8f4712b` — every commit above is at or before this sha; nothing landed after it unreviewed.

- Gate pass (head c8a7c52), 1 medium: the fail-loud guard trusted `source`'s exit status, so a lib missing some `worker_alert_*` functions let both hooks exit 0 with no `not-sent` line. **Fixed** in c7511ea: `declare -F` on every function each hook calls, missing symbol routed through `lib_missing()` and named; one incomplete-lib test per suite.
- Second pass (head c7511ea), 1 medium: `lib_missing()` ignores a failed `mkdir` or append (read-only home, full disk). **Disputed**: an unwritable `~/.claude` breaks the harness's transcript and session registry before any hook runs, a Stop hook's stderr is not a durable channel, and the same posture predates this PR in both hooks' normal log writes. Not worth a third round.


caneff:
## Codex adversarial pass (gate, head c8a7c52)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new fail-loud guard only detects an un-sourceable library, so a sourceable but incomplete/version-skewed helper again makes both hooks silently lose alerts.

Findings:
- [medium] A sourceable incomplete helper bypasses the fail-loud path (flow/claude/hooks/worker-stop-alert.sh:40-45)
  `source` returning zero proves only that the file parsed and ran successfully; it does not prove the required helper functions were defined. A valid stale/truncated library can therefore pass this branch, after which calls such as `worker_alert_resolve_session` and `worker_alert_logline` become command-not-found failures. Neither hook uses `set -e`, so it continues to exit 0; critically, the missing `worker_alert_logline` also prevents the required `not-sent` record. This recreates the silent-loss failure for a realistic partial/version-skew deployment, while the new tests cover only an absent file.
  Recommendation: After sourcing, verify every required function with `declare -F` (at least read_brief, resolve_session, logline, worker-agent-name, and controller-pane). Treat any missing symbol as `lib-missing` using a logging path that does not depend on the library; add the same sourceable-incomplete-library case to both hook suites.

Next steps:
- Add helper-interface validation and regression tests before shipping.


Controller: confirmed against the source; one fix round sent to the worker.

caneff:
## Codex adversarial pass (second, head c7511ea)

# Codex Adversarial Review

Verdict: needs-attention

No ship: the promised fail-loud helper-missing path can still exit 0 without a durable alert.

Findings:
- [medium] lib-missing logging failures are silently ignored (flow/claude/hooks/worker-stop-alert.sh:37-39)
  If ~/.claude cannot be created or the alert log cannot be opened (read-only home, full disk, or a colliding path), both commands fail but the function unconditionally exits 0. The hook therefore loses the only required lib-missing record precisely in a degraded environment, recreating the silent no-controller-resolution failure this change is meant to expose.
  Recommendation: Check the mkdir and append results; on failure emit a lib-missing diagnostic to stderr before exiting 0. Apply the same change to the spin hook and add a test with an unwritable or non-file log target.

Next steps:
- Harden lib_missing logging and rerun both hook suites in a writable test environment.

PR #1067, opened 2026-09-22
loop.py announce: resolve each worker's herdr name at send time

### LINKED ISSUE #1013: loop.py announce hands the worker's herdr agent name straight to the sender
`burndown/loop.py` `announce` (grep `entry["agent"]`, the resume re-announce path) passes each live worker's herdr agent name unchanged to the send callback. The run file keys workers by herdr name on purpose (#892, #923): it survives a restart, and it is not an address. SendMessage takes a live Claude session name, so a resumed controller's re-announcement never reaches a worker unless the caller resolves the name first — and `loop.py`'s own test passes a permissive callback that accepts anything, so nothing witnesses a resolved recipient. Raised by PR #1012's final Codex pass.

Fix: resolve each herdr name to the current Claude session name at the sending boundary, immediately before the send, through `herdr agent list` and the session registry (the same resolution `implement-dispatch` does for the controller after #923); refuse, naming the worker, when it does not resolve. Test with a callback that rejects the durable name and accepts only the resolved session name.

Filed from PR #1012's final Codex pass (controller disposition), 2026-09-21.

## Blocked by

caneff:
Closes #1013

## What changed

- `loop.announce` now resolves each worker's herdr agent name to the live Claude session `send` can reach, immediately before each send, via a new `resolve` argument (default `resolve_via_binary`, which shells out to `resolve-controller <name>`).
- A name that does not resolve refuses by name, naming the worker, rather than being passed through to `send`.
- `references/loop.md` § What resume owes each worker updated to describe the resolve step happening inside `announce` itself.

## Tests run

`python3 burndown/loop_test.py` — 102 passed.

## Decisions made

Round 1 (standards/spec/correctness, 10 findings) and one verification pass, both confirmed. Every finding below cites its sidecar id.

- S1 — fixed (1d0767b): extracted `stubbed_resolve_controller`, replacing three copies of the PATH-stub ritual.
- S2 — fixed (1d0767b): the same helper uses `os.environ.get("PATH", "")`, so a caller with no `PATH` set no longer raises.
- S3 — fixed (1d0767b): trimmed `announce`'s docstring, which had grown to restate `references/loop.md` near-verbatim and narrate its own test.
- S4 — disputed: `burndown/SKILL.md` and `references/run-file.md` describe the controller performing the resolve — accurate in production, since `announce` has no CLI subcommand (`loop.py`'s own parser text: "sending is `SendMessage`, which the controller calls itself") and is never invoked outside tests. `references/loop.md`'s edit describes `loop.announce`'s own model of the same algorithm; no contradiction.
- P1 — disputed: same underlying issue and reasoning as S4.
- P2 — fixed (1d0767b): added `test_announce_uses_the_default_resolver_when_none_is_passed`, calling `announce(state, send)` with no third argument over a stubbed binary, so the default-resolver wiring has its own witness.
- C1 — fixed (1d0767b): `resolve_via_binary`'s two refusal guards (non-zero exit, empty stdout) used to mask each other under mutation — deleting either alone left the suite green. Two new stub cases and tests isolate each.
- C2 — disputed: the refusal text only promises that already-reached workers are not re-sent on retry, never that a failed resolve succeeds next time. Stop-on-first-failure predates this diff (the same shape already existed for `send` failures); out of scope for #1013.
- C3 — fixed (1d0767b): same helper as S2 fixes the `None`-`PATH` case, and its `finally` around the block means a failing test body no longer leaks `PATH` globally.
- C4 — disputed: `resolve_via_binary` has no production call site because `announce` itself has none — a pre-existing structural property (`SendMessage` is a tool call, not something a Python module can invoke), unrelated to this diff and out of #1013's scope, which asked only for `announce`'s own contract and test.

## Last reviewed sha

1d0767b9a5ae — no commits past it.

- Gate pass (head 1d0767b): approve, no material findings. No second pass.


caneff:
## Codex adversarial pass (gate, head 1d0767b)

# Codex Adversarial Review

Verdict: approve

Ship: the durable herdr name is resolved at the send boundary, resolution failures refuse by worker name, and tests now prove the send callback receives only the resolved session name.

No material findings.

PR #1068, opened 2026-09-22
Cover replayed commits in the commit-identity guard

### LINKED ISSUE #1006: Cover replayed commits in the commit-identity guard
The #934 commit-identity guard is a pre-commit hook (`flow/lane/hooks/commit-identity-guard.sh`), so it does not fire on rebase, cherry-pick or merge. Probed: `git -c user.email=bad@x.com rebase <base>` rewrote the committer email of the replayed commit with exit 0, and the lane mandates a rebase onto main before the PR — the same email-privacy push rejection as #909, reached through a required operation.

Needs a mechanism covering replayed commits — e.g. a pre-push check of the pushed commits' author/committer emails against the checkout's `user.email`, or a post-rewrite hook — installed by `implement-dispatch` alongside the guard, with the same `COMMIT_IDENTITY_OVERRIDE` escape.

Filed from #934 review finding C1 (map #776).

## Blocked by

- #934

caneff:
Closes #1006

## What changed

- Installs a second commit-identity-guard hook as `pre-push`, checking every commit's author/committer email in the pushed range against the checkout's configured `user.email`, with the same `COMMIT_IDENTITY_OVERRIDE` escape as the pre-commit half — this is what catches a rebase or cherry-pick, which never goes through `git commit` at all.
- Generalizes the #1009 pre-commit hook-takeover machinery (byte-identity ownership, foreign-hook preservation) to install both hook slots from one small table.
- `flow/claude/CLAUDE.md` and `flow/claude/OPERATIONS.md` updated to describe both hook slots instead of only pre-commit.

## Tests run

`bash tests/all.sh` (78 suites, including the new `flow/lane/hooks/commit-identity-guard-pre-push.test.sh` and the pre-push cases in `flow/lane/tests/implement_dispatch.rs`) — PASS, run four times across the build (before/after the round-1 fixes and again after the post-verification leak fix).

## Decisions made

Round-1 review (multi-axis-code-review, 3 parallel axes against 77cfb18): 11 findings, all fixed in 4586ef6.

- S1 — fixed. A foreign pre-push hook that reads stdin (git's own ref-list protocol) drained it before the guard's own read, so the guard checked nothing and exited 0. Buffered stdin to a temp file so both the foreign hook and the guard see the full ref list.
- S2 — fixed. Zero ref lines read on stdin fell through to a silent exit 0; now refuses and names itself.
- S3 — fixed. The new-branch commit range hard-coded `--remotes=origin`; now reads the hook's own `$1` (the remote git actually passes it).
- S4 — fixed. `install_hook_slot`'s five positional params (a slot data clump) replaced with a small `(slot, foreign_name, guard_name, guard_content, buffer_stdin)` table `install_identity_guard` iterates.
- P1 — fixed. The shared wrapper had grown `"$@"` on the guard's `exec` line for no functional reason, changing pre-commit's wrapper bytes; since ownership is byte-identity, this read every already-dispatched repo's own wrapper as foreign on the next dispatch. Restored pre-commit's wrapper to its exact pre-#1006 bytes.
- P2 — fixed. `flow/claude/OPERATIONS.md` updated to describe the pre-push slot.
- C1 — fixed. Same root cause and fix as S1 (duplicate finding from the correctness axis, reproduced concretely with a stdin-draining foreign hook).
- C2 — fixed. Same root cause and fix as P1 — the changed wrapper bytes would displace an already-installed wrapper into `pre-commit.foreign`, whose own body hard-codes that same path, causing self-recursion. Added a regression test pinning the pre-#1006 wrapper bytes and asserting no displacement on redispatch.
- C3 — fixed. The shell test's "no configured user.email is refused" case only asserted exit 1, which also fires from the unrelated email-mismatch branch. Added an assertion on the specific refusal message.
- C4 — fixed. The Rust test's foreign pre-push hook fixture was `exit 0` (never touched stdin), so it could never have caught S1/C1. Now drains stdin like a real hook, and the test asserts both the specific mismatched email is named and a correctly-configured push through the same hook still succeeds.
- C5 — fixed (doc-only). Checking both AUTHOR and COMMITTER email is a real, expected refusal on an imported/co-authored commit; documented `COMMIT_IDENTITY_OVERRIDE` as the escape.

Verification pass (against the 4586ef6 fix diff): all 11 confirmed fixed, with its own mutation checks on the two hard fixes. It surfaced one new, minor observation not itself a round-1 finding: the buffered wrapper's `exec` of the guard skipped its own `trap ... EXIT` cleanup, leaking one `mktemp` file per push. Fixed directly in 5ef2ad1 (dropped the `exec`; the guard's exit status still propagates as the wrapper's own since it's the last command) rather than filed as a follow-up — trivial, fully covered by a new leak-regression test, and safe to include given the full gate re-ran green afterward.

Codex gate (against 5ef2ad1): one `[high]` — `cat >"$stdin_buf"` (the pre-push wrapper's buffering copy) ignored its own exit status; a copy that fails partway (disk full, an I/O error) after writing one or more complete ref lines hands both the foreign hook and the guard a truncated-but-nonempty ref list, so the zero-ref refusal (S2) never fires and a replayed commit on the ref that got cut off would push through unchecked. Fixed in e607eb2: the wrapper now refuses, naming itself, before either hook runs when the copy fails — with a regression test that stubs `cat` on PATH to write one line and exit 1, and asserts the push is refused and neither hook's marker file was written, mutation-checked against the unfixed line.

## Last reviewed sha

4586ef622fb3bdfff9c7e3e150b178d70b462fa9 — three commits past it were not re-reviewed by the axes or the verification pass:
- 5ef2ad1 — test-only, the stdin-buffer leak fix above, covered by its own new regression test.
- e607eb2 — the Codex-gate fix above, covered by its own new regression test; Codex's own second pass (controller-run) is what re-reviews this commit.

- Gate pass (head 5ef2ad1), 1 high: the buffered pre-push wrapper ignored `cat >"$stdin_buf"`'s exit status, so a truncated ref list could pass with refs unchecked. **Fixed** in e607eb2 (copy failure refuses before either hook runs; stub-`cat` regression test, mutation-checked).
- Second pass (head e607eb2), 1 high + 1 medium, both in the pre-push guard's ref loop: the 40-char zero sentinel refuses new-branch and deletion pushes in a SHA-256 repository (fail-closed, none of ours), and `saw_a_ref` is set before the record is validated so a malformed line exits 0 unchecked. **Filed** as #1069, blocked by #1006: no third round.


caneff:
## Codex adversarial pass (gate, head 5ef2ad1)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: a temp-file write failure can turn a partially buffered pre-push ref stream into an approved push, bypassing the replayed-commit identity check.

Findings:
- [high] Partial stdin buffering can silently omit pushed refs (flow/lane/src/bin/implement_dispatch.rs:377)
  The wrapper ignores `cat >"$stdin_buf"`'s exit status. If the temp filesystem fills or errors after at least one complete ref line is written, the foreign hook and guard receive a nonempty but truncated list. The guard therefore skips the omitted ref(s), does not trigger its zero-ref refusal, and can return success after checking only matching commits. A multi-ref push can consequently send a replayed foreign-email commit on an omitted ref—the exact privacy rejection this guard is meant to prevent.
  Recommendation: Make buffering fail closed: check the `cat` status and immediately emit a refusal and exit nonzero on any read/write failure before invoking either hook. Add an end-to-end wrapper test that simulates a partial buffer write and proves the push is refused rather than approved.

Next steps:
- Fail closed on stdin-buffer copy failure and add a regression test.


Controller: confirmed against the source; one fix round sent to the worker.

caneff:
## Codex adversarial pass (second, head e607eb2)

# Codex Adversarial Review

Verdict: needs-attention

No ship: the guard breaks normal first pushes in SHA-256 repositories, a Git-supported format.

Findings:
- [high] SHA-256 repositories cannot push new branches or delete refs (flow/lane/hooks/commit-identity-guard-pre-push.sh:34-42)
  The hook recognizes only the 40-character SHA-1 all-zero sentinel. Git's pre-push protocol uses a 64-character zero OID in SHA-256 repositories, so a new-branch push is treated as an update and `git rev-list <64-zero>..<local>` fails; deletion is similarly misclassified. The guard then refuses the push. This blocks the normal first PR push in any SHA-256 repo.
  Recommendation: Derive the null OID from the received OID length or the repository object format, and add real pre-push coverage using `git init --object-format=sha256` for both a new branch and a deletion.
- [medium] Malformed pre-push input is accepted as a clean check (flow/lane/hooks/commit-identity-guard-pre-push.sh:38-42)
  `saw_a_ref` is set before the record is validated. A blank or one-field line therefore satisfies the nonempty-input fallback, then the missing `local_sha` is silently skipped and the hook exits 0 without checking any commit. This contradicts the hook's fail-closed handling of abnormal invocation and leaves malformed ref input indistinguishable from a clean push.
  Recommendation: Require exactly four nonempty protocol fields for non-deletion updates, reject malformed records explicitly, and add regression cases for blank and truncated records.

Next steps:
- Fix the object-format-dependent null OID handling and add SHA-256 hook coverage before merging.

PR #1071, opened 2026-09-22
Route the herdr toast install step from flow/install.sh

### LINKED ISSUE #1016: Call herdr-toast-install from flow/install.sh

caneff:
Closes #1016

## What changed

- `flow/install.sh` now routes the toast step through `flow/bin/herdr-toast-install`, after `backup-sync.sh --restore` and before `lane-install.sh`, since it is the one step here that reaches outside the filesystem (the Windows registry) and is the most likely to fail on a given machine.
- Only the installer's two by-design skips (a linked worktree, a path under /tmp) are swallowed as "skipped"; every other failure — including its third refusal, a real directory at a link destination — is a real machine problem: the rest of the install still runs, but install.sh exits nonzero at the end and names the toast step.
- `flow/install.test.sh` asserts the installer's own refusal text alongside install.sh's summary line, stubs `reg.exe`/`powershell.exe`/`wslpath` so no run in the file can reach the real registry, adds a success-path case (via `HERDR_TOAST_ALLOW_TMP=1`) that exercises the real link-and-register behavior, pins `TMPDIR=/tmp` so the refusal case doesn't depend on the caller's environment, and covers both the non-refusal-failure exit code and the real-directory refusal not being swallowed.

## Tests run

`bash tests/all.sh` — 77 suites passed (one transient, unrelated `tests/lane-unparked.test.sh` failure on an earlier run did not reproduce on rerun).
`bash flow/install.test.sh` — ALL PASS (24/24).
`bash flow/bin/herdr-toast-install.test.sh` — ok (unmodified by this diff).

## Decisions made

Every round-1 finding was fixed; none disputed, filed, or handed back.

- S1 (hard) — fixed, `cd45202`: distinguished the installer's by-design refusal from any other failure.
- S2 (judgement) — fixed, `cd45202`: moved the step after `backup-sync.sh --restore`.
- S3 (judgement) — fixed, `cd45202`: rewrote the duplicated rationale comment.
- P1 (judgement) — fixed, `cd45202`: strengthened the test to also check the installer's own refusal text.
- P2 (judgement) — fixed, `cd45202`: added the `HERDR_TOAST_ALLOW_TMP=1` success-path case.
- C1 (hard) — fixed, `cd45202`: same mechanism as S1.
- C2 (judgement) — fixed, `cd45202`: stubbed `reg.exe`/`powershell.exe`/`wslpath` unconditionally.
- C3 (judgement) — fixed, `cd45202`: same success-path case as P2.
- V1 (verification residual, judgement) — fixed, `b21f60f`: pinned `TMPDIR=/tmp` so the refusal case doesn't depend on the caller's own `TMPDIR`.
- Codex gate finding 1 (high) — fixed, `bed18e8`: install.sh now exits nonzero and names the toast step when it fails for a non-refusal reason, instead of exiting 0.
- Codex gate finding 2 (medium) — fixed, `bed18e8`: narrowed the skip match to the installer's two by-design skips only; its real-directory refusal now goes down the failure path.

## Last reviewed sha

bed18e8ea1de3d8f785f0ffe27a65a1ad692f965 — no commits past this sha.

- Gate pass (head b21f60f), 1 high + 1 medium: a real toast-install failure still exited 0, and the "real directory" refusal was swallowed as a by-design skip. **Fixed** in bed18e8 (install finishes its steps then exits 1 naming the toast step; skip matched on the two exact by-design texts; four new tests).
- Second pass (head bed18e8): approve, no material findings.


caneff:
## Codex adversarial pass (gate, head b21f60f)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: genuine toast-install failures are reported but converted to a successful overall install, and a destination conflict is also misclassified as an allowed skip.

Findings:
- [high] Real toast installer failures do not fail the install (flow/install.sh:70-77)
  The `if !` consumes the toast command’s nonzero status. In the non-refusal branch the script only prints an error, then continues to `lane-install.sh`; if that succeeds, `install.sh` exits 0 and prints `Done`. A failed registry write, missing `wslpath`, or failed link operation therefore leaves the handler/scripts absent while automation and a human caller see a successful installation. This contradicts the requirement not to hide non-refusal failures.
  Recommendation: After printing a non-refusal failure, exit nonzero (or preserve and return the toast command’s status). Add a test that makes the toast command fail without an allowed refusal and asserts `install.sh` fails while prior install steps remain completed.
- [medium] Prefix matching swallows a real destination conflict as an allowed refusal (flow/install.sh:70-73)
  The allowed-skip check accepts any output line beginning `refusing: `. But `herdr-toast-install` emits exactly that prefix when a target such as `~/.local/bin/herdr-focus-latest` is a real directory. That is not either of the controller-approved disposable-checkout cases; it prevents the required toast links from being installed, yet this code reports a benign skip and exits successfully. The same broad match can mask any future failure that emits a similarly prefixed diagnostic.
  Recommendation: Recognize only the linked-worktree and `/tmp` refusal messages (preferably via a distinct exit status or machine-readable signal from `herdr-toast-install`); propagate all other failures. Add coverage for a real destination directory and require a nonzero install result.

Next steps:
- Preserve non-refusal toast failure status and narrow the skip classification to the two explicitly allowed refusal conditions.


Controller: both confirmed against the source; one fix round sent to the worker (finish the install, exit nonzero on a real toast failure; narrow the skip to the two by-design refusals).

caneff:
## Codex adversarial pass (second, head bed18e8)

# Codex Adversarial Review

Verdict: approve

Ship: the installer invokes the toast step, permits only the two intended refusals while continuing, and preserves a nonzero outcome for all other toast failures.

No material findings.

PR #1072, opened 2026-09-22
Fix notify-send hang when a value flag is the last argument

### LINKED ISSUE #1017: notify-send hangs when a value flag is the last argument

caneff:
Closes #1017

## What changed

- `flow/bin/notify-send`: a value-taking flag as the last argument left `shift 2` with only one positional to consume, spinning the parse loop forever. Drop the flag when no value follows instead. Also adds a `NOTIFY_SEND_WSCRIPT` seam so the real Windows `wscript.exe` call can be redirected in tests.
- `flow/bin/notify-send.test.sh`: regression test running each value-taking flag last under a `timeout`, pointed at a no-op `wscript.exe` stub via `NOTIFY_SEND_WSCRIPT` so it never fires a real toast.
- Test hardened after round-1 review, then again after the Codex gate pass: real rc-0 witness instead of "not 124", a flag-list-vs-parser-case-arm drift check, and (Codex finding) the earlier mount-namespace isolation and its capability skip replaced outright with the `NOTIFY_SEND_WSCRIPT` seam, since that skip could read green on a host where the parser never ran.

## Tests run

`bash tests/all.sh` — 79 suites passed (run twice at the final sha to rule out an unrelated flake in `tests/lane-unparked.test.sh` seen once, not reproduced, not touched by this diff), including `flow/bin/notify-send.test.sh`.

## Decisions made

Round 1 (multi-axis review, head b7db1ef):
- S1 (hard): fixed — rc now matched 0/124/97/other explicitly instead of only checking for 124.
- S2 (hard): fixed — a failed `mount --bind` now failed the case loud (exit 97) instead of being swallowed by `2>/dev/null`. (This mechanism was later replaced outright — see Codex gate below.)
- S3 (judgement): fixed — the test's flag list is diffed against the parser's own case arm so the two can't drift apart silently. Still in place.
- P1 (judgement): disputed — the ticket asks to fix the hang, not to add real notify-send's error-and-exit behavior for a dangling flag; the shim already silently drops other unrecognized/malformed input, so silently dropping this one matches its existing style.
- P2 (judgement): fixed — same fix as S1/C1.
- C1 (hard): fixed — same fix as S1, plus a real `unshare -rm true` capability probe replacing `command -v unshare`. (Probe later removed outright — see Codex gate below.)
- C2 (hard): fixed — same fix as S2.
- C3 (judgement): disputed — same reasoning as P1.

All 8 round-1 findings confirmed by an independent verification pass (mutation-tested against fix commit b7db1ef); no disagreements with the dispositions above.

Codex gate pass (head b7db1ef): one medium — the unshare-refused SKIP path exited 0, and `tests/all.sh` reads exit 0 as PASS, so the gate could read green on a host where the regression test never ran (the absent-run-as-success shape `docs/agents/defect-classes.md` rules against). Fixed in 3f438ee: dropped the mount-namespace isolation and the capability-dependent skip entirely, replaced with a `NOTIFY_SEND_WSCRIPT` env seam on the shim itself so the test always runs, everywhere, with no skip path. Mutation check: reverted the shift fix and saw every flag case time out (rc 124).

## Last reviewed sha

3f438ee — no commits past this sha.

- Gate pass (head b7db1ef), 1 medium: the test exited 0 where `unshare -rm` was refused, an absent run reading as a pass. **Fixed** in 3f438ee by controller ruling: no isolation and no skip; the shim reads `NOTIFY_SEND_WSCRIPT` and the test points it at a stub, so the test always runs.
- Second pass (head 3f438ee), 1 medium: unchecked `mktemp`/stub write/`chmod` in test setup. **Disputed**: the regression witness is the timeout on the parse loop (mutation: 16 cases hit rc 124), which a missing stub does not affect; a failed setup fires a real toast at worst, not a false pass. Not worth a third round.


caneff:
## Codex adversarial pass (gate, head b7db1ef)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new regression suite can report success without exercising the parser at all.

Findings:
- [medium] Namespace refusal turns zero test coverage into a passing gate (flow/bin/notify-send.test.sh:30-33)
  When `unshare -rm` is unavailable/refused, the suite exits 0 before any flag case runs. `tests/all.sh` treats an exit-0 suite as PASS, so the end-to-end gate can be green precisely in environments where this regression test did not execute. This recreates the repo's documented absent-run-as-success defect class and leaves the hang unguarded on such runners.
  Recommendation: Make isolation available without an optional user namespace (for example, introduce a test-only executable/launch-path seam), or make unavailable isolation fail the merge-gate suite rather than exit 0; retain an explicit skip only outside required CI coverage.


Controller: confirmed; one fix round sent (replace the namespace isolation and its skip with a NOTIFY_SEND_WSCRIPT seam so the test always runs).

caneff:
## Codex adversarial pass (second, head 3f438ee)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new regression test can pass after its required stub setup fails.

Findings:
- [medium] Stub-setup failure is treated as a passing test run (flow/bin/notify-send.test.sh:25-29)
  `mktemp`, the stub write, and `chmod` are unchecked. With `set -u` but no `set -e`, a failed `mktemp` leaves `tmp` empty; the test then passes a nonexistent `/wscript-stub.sh`. The shim backgrounds that failed launch and exits 0, so every flag reports PASS. This was reproducible in the provided read-only environment: mktemp/write/chmod failed, followed by ALL PASS. That makes a failed test setup indistinguishable from a completed isolated run.
  Recommendation: Fail immediately if `mktemp`, stub creation, or `chmod` fails (for example, guard each command with `|| exit 1`). Also make the stub invocation observable if the test intends to claim the parser completed through the seam.

Next steps:
- Harden test setup and rerun the regression suite.

Issue #1028, opened 2026-09-22
implement: Codex second-pass small findings fixed in-round; third run only after a fixed high


caneff:
## Parent

#1024

## TL;DR

When the final Codex review finds a small problem, the worker fixes it instead of a new ticket being filed, and there is a hard cap on how many Codex reviews one PR can get.

## What to build

In `implement`'s merge step, a second-pass Codex finding that meets the adjacent-fix rule (#1025) is fixed by the worker in one round and the controller reads the fix diff itself. A third Codex run happens only when the fixed second-pass finding was high, and the third run is final: its findings are `disputed` or `leftover`, never a fourth run. The controller still evaluates every recommendation before it reaches the worker. Grill rulings Q2, Q14 on #1024.

## Acceptance criteria

- [ ] The merge step says which second-pass findings the worker may fix and that the controller reads the diff.
- [ ] The third-run condition and its finality are stated in one sentence each.
- [ ] `codex-pass-schedule.test.sh` covers the new sentences and fails without them.
- [ ] `tests/all.sh` green.

## Blocked by

- #1025

## Seams under test

- The codex-pass-schedule wording test over `implement`'s merge step.

PR #1081, opened 2026-09-22
burndown: box check counts herdr's working agents, not every claude process

### LINKED ISSUE #1075: loop.py box check: count working agents, not every Claude process
`burndown/loop.py`'s box check (`count_agent_processes`, `box_check`) counts every process named `claude` on the box against the 28 cap. An idle session costs no cores: measured 2026-09-22 10:30, the box held 15 `claude` processes of which herdr reported 6 `working` and 9 `idle`/`done`; the idle ones sat at 1–2% CPU and ~450 MB each. The cap is Chris's core rule ("no more than 28 cores TOTAL"), so counting the idle ones held run `burn-2026-09-22-0636` to two and three live workers all morning and gave its spec run 2 slots instead of 3. Chris, 2026-09-22: "i dont think stale and idle ones should count", then "change loop.py".

Change: the count is the agents `herdr agent list` reports with `agent_status` `working` (read the JSON, not the terminal titles), plus the workers the tick is about to start at their peak, as today. A process count stays as the fallback only when herdr cannot answer, and the refusal line names which counter it used. `--processes <n>` keeps overriding both. The `peak:` line prints the working count and the total count side by side so a controller can see what was excluded. Subagents of a working worker are separate `claude` processes that herdr lists as their own working agents, so the peak reserve per slot is unchanged. Tests: a fake herdr listing with mixed statuses, the fallback when the listing fails, and the override.

Filed from conversation (controller skills-dc, run burn-2026-09-22-0636), 2026-09-22.

## Blocked by

### LINKED ISSUE #1079: loop.py: zero working herdr agents reads as an empty box
`burndown/loop.py`'s `count_working_herdr_agents` (added by #1075) raises on a failed herdr call, non-JSON output, a missing `agents` key, or a non-list `agents`, but returns `0` for a well-formed listing with no `working` entry, e.g. `{"result":{"agents":[]}}`. `agent_count` then reads that `0` as a fully-idle, wide-open box and dispatches a full tick onto a box that may actually be loaded.

`count_agent_processes` already guards the equivalent case for `ps`: a healthy listing with zero `claude` processes is refused rather than read as zero agents, because the controller running the tick is itself a `claude` session, so a genuine zero is impossible. The same argument holds for herdr — the controller running the tick is itself `working` — but `count_working_herdr_agents` has no such floor.

Fix: raise `LoopError` in `count_working_herdr_agents` when the working count is `0`, mirroring the `ps` path, so `agent_count` falls back to the process count and names the fallback in the refusal. Add a test covering `{"result":{"agents":[]}}` to `burndown/loop_test.py`.

Confirmed via mutation and a live measurement by the #1075 correctness review (`~/.cache/agent-reviews/skills/review-correctness-1075.md`, finding C1):

```
loop.agent_count(Args, herdr=(0,'{"result": {"agents": []}}'), ps=<16 claude>) -> (0, HERDR_COUNTER, 16)
loop.box_check(0, 0, counter=HERDR_COUNTER) -> {'ok': True, 'refusals': []}
```

Filed rather than fixed in-round on #1075's own PR: the fix needs a new regression test in `burndown/loop_test.py`, a second file, which breaks the adjacent-fix rule's one-file bound (`multi-axis-code-review/check_adjacent.py` requires exactly one file changed).

Filed from #1075's implement-lane review (controller skills-dc).

## Blocked by

caneff:
Closes #1075
Closes #1079

## What changed

- `burndown/loop.py`'s box check now reads `herdr agent list`'s working panes, matched to pids through the sessions registry — validated against `/proc/<pid>/stat`'s own start time, since a stale record can survive a WSL restart or an ordinary pid reuse — plus every `claude` pid no valid match resolves to (a subagent or headless run). A pane counts as working unless its `agent_status` is exactly `idle` or `done` (a missing/unrecognised status fails closed as working too). It falls back to the raw `ps -eo comm=` process count only when herdr or the pid match cannot be taken at all.
- `--processes <n>` still overrides both; a refusal names which counter gated it, and the `peak:` line prints the working/unlisted split beside the total.
- `burndown/SKILL.md` and `burndown/references/loop.md` updated to match.

## Tests run

`python3 burndown/loop_test.py` — 119 passed.
`bash tests/all.sh` — 85 suites passed.

## Decisions made

- S1 (standards, judgement) — a well-formed herdr listing with zero `working` entries reads as an empty box: **fixed**, sha `7a6bedd`, closing #1079.
- S2 (standards, judgement) — counter-label drift: **fixed**, sha `08e7e88` plus `7a6bedd`/`9391bd8`/`a70db34` (wording kept current through every mechanism change).
- S3 (standards, judgement, duplicated code) — duplicated default subprocess runner: **fixed**, sha `08e7e88`.
- P1 (spec, judgement) — same bug as C1: **fixed**, sha `7a6bedd`, closing #1079.
- P2 (spec, judgement) — the `peak:` line omits the total and the counter used on the fallback path: **leftover**.
- C1 (correctness, CONFIRMED/high) — zero-working-agents read as an empty box: **fixed**, sha `7a6bedd`, closing #1079.
- C2 (correctness, judgement) — a missing herdr falls back silently: **leftover** (same gap as P2).
- C3 (correctness, judgement) — the original docstring claimed a subagent is its own working herdr entry: **fixed**, sha `08e7e88`, superseded by `7a6bedd`'s mechanism fix.
- Codex gate finding #1 (7a6bedd's own review) — a resolved pane's `agent_status` was matched against the literal `"working"`, dropping a missing/unrecognised status out of the count entirely: **fixed**, sha `9391bd8`.
- Codex gate finding #2 (9391bd8's own review) — `_session_pids` joined a registry record to a pid without checking the record still belongs to that process, so a stale record (pid reuse, WSL restart) could silently absorb a live unlisted claude process's pid: **fixed**, sha `a70db34` (validated against `/proc/<pid>/stat` field 22, the same comparison `resolve-controller` and `worker-alert-lib.sh` already make).

Verification pass (round 1, before the controller ruling): PASS — full report and dispositions at `~/.cache/agent-reviews/skills/review-verify-1075.md` / `dispositions-1075.jsonl`. `7a6bedd`, `9391bd8` and `a70db34` were controller/Codex-directed fixes on top, each covered by `tests/all.sh` and by-hand mutation checks rather than a separate multi-axis round.

## Last reviewed sha

`08e7e88` (multi-axis review, round 1 + verification). `7a6bedd`, `9391bd8` and `a70db34` are controller-ruling and Codex-gate-directed fixes on top, each checked by `tests/all.sh` (85 suites) and the mutation checks noted below.

## Mutation checks (this session)

On `7a6bedd`: disabling the herdr-listing-matches-no-pids refusal (caught, restored); dropping the unresolved-pane fail-closed branch (caught, restored).
On `9391bd8`: restoring the `agent_status == "working"` equality test (caught, restored).
On `a70db34`: dropping the `procStart` comparison in `_session_pids` (caught by both the unit test and the integration-level stale-record test, restored).

- Gate pass (head 7a6bedd), 1 high: a resolved pane with a missing or unknown status was neither working nor unlisted. **Fixed** in 9391bd8 (only explicit idle/done exclude).
- Second pass (head 9391bd8), 1 high: the sessionId-to-pid join trusted a stale registry record, so a reused pid could be absorbed by an idle pane. **Fixed** in-round in a70db34 (`procStart` validated against `/proc/<pid>/stat`); controller read the fix diff at a70db34.
- Third pass (head a70db34, final), 1 high: a foreign controller's working pane gets no fan-out reservation, so two runs on one box can pass the cap together. **Leftover** (CX3 in `dispositions-1075.jsonl`): pre-existing under the raw process count, and a box-wide reservation is a design of its own; the sweep harvests it.
- Before the ruling round: #1079 folded in and closed by this PR.


caneff:
## Codex adversarial pass (gate, head 7a6bedd)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: malformed or newly introduced herdr statuses can be read as idle and disable the shared-core cap.

Findings:
- [high] Unknown agent_status is silently excluded from the cap (burndown/loop.py:378-389)
  A listed pane with a resolvable current Claude PID is added to matched_pids, but only the exact string "working" increases working. Thus a valid JSON response with a missing, null, or future/unknown agent_status produces (0, 0) for a sole controller pane: it is neither working nor unlisted. agent_count then supplies zero to box_check and can dispatch a full tick on a loaded box. This violates the fail-closed invariant and treats schema drift/partial data as benign.
  Recommendation: Validate each resolved pane status. Count only explicitly idle/done panes as excluded; count missing or unrecognized statuses as working (or reject the herdr result and use the conservative process-count fallback). Add a regression test for a resolvable pane with absent and unknown agent_status asserting it cannot yield a zero count.

Next steps:
- Fix the status handling and run the loop test suite.


Controller: confirmed; one fix round sent (exclude only explicit idle/done, everything else counts as working).

caneff:
## Codex adversarial pass (second, head 9391bd8)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: stale session-registry records can make an active unlisted Claude process appear to be an idle/done pane, undercounting the shared-box cap.

Findings:
- [high] PID reuse lets a stale idle pane exclude a live Claude process (burndown/loop.py:283-307)
  The session join trusts `sessionId -> pid` without validating that the registry record still belongs to that process. If an old idle/done pane's stale record points at a PID that has been reused by a new unlisted subagent/headless Claude process, that PID enters `matched_pids` and is excluded from `unlisted`; the idle pane adds nothing to `working`. For example, an idle stale pane mapped to PID 100 plus a real working pane at PID 101 yields a count of 1 for two active Claude processes. This defeats the 28-core safety cap. The repo's existing session reader explicitly checks `procStart` because PID reuse is expected, especially after WSL restarts.
  Recommendation: Validate each registry record against the live PID and its recorded `procStart` before adding it to the session map, and only treat matches intersecting `claude_pids` as resolved. Add a regression test with a stale idle record whose reused PID is an otherwise-unlisted live Claude process; it must remain counted or force the flat-process fallback.

Next steps:
- Add the stale-PID regression test and procStart validation, then rerun the box-count tests.


Controller: confirmed; fixed in-round under § The merge as landed by #1028 (a second-pass finding is fixable; a fixed high earns a third run).

caneff:
## Codex adversarial pass (third, head a70db34)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: the revised counter still permits a shared-box peak to exceed the 28-core cap when another run's working pane fans out after this tick measures it.

Findings:
- [high] Foreign working panes have no peak reservation (burndown/loop.py:452-456)
  The measured count includes each foreign working pane as only one process, and `projected_processes` reserves fan-out headroom only for this run's `live` workers. A foreign worker can therefore be counted as one immediately before it launches its own review subagents; those processes did not exist at the `ps` snapshot and are neither `unlisted` nor reserved. For example, a measured 23 that includes one such foreign worker permits one local slot (projected 28); the foreign worker then adding its four-process fan-out drives the box to 32. This violates the stated total-core safety rule on precisely the shared-box concurrency the gate is meant to protect.
  Recommendation: Use a box-wide reservation mechanism, or conservatively reserve remaining fan-out headroom for every herdr working pane that can launch it, accounting for already-observed subagents so they are not double charged. Add a regression test where a foreign working pane has no currently visible subagents but may fan out after the measurement.

Next steps:
- Block dispatch until the peak calculation covers concurrent foreign-worker fan-out.

Issue #1087, opened 2026-09-22
flow/lane: neither call site of canonical_workspace_path is witnessed


caneff:
`flow/lane/src/workers.rs::canonical_workspace_path` (added in #1040) is
unit-tested directly, but neither of its two call sites is: reverting
`implement_dispatch.rs`'s `WorkerRecord { workspace: .. }` construction and
`merge_cleanup.rs`'s `remove_workspace` call back to their pre-#1040 raw
`.display().to_string()` form leaves the full crate `cargo test` green —
only the helper is pinned, nothing pins that either binary actually calls
it. That's the exact defect class #1040 exists to prevent, reopened by
whichever binary next touches either call site.

Confirmed by the correctness axis of #1040's `/multi-axis-code-review`
round, 2026-09-22 (finding C1): reverting both call sites and running
`cd flow/lane && cargo test` produced zero failures.

Fix direction: an integration-level test exercising `implement-dispatch`'s
record write and `merge-cleanup`'s `remove_workspace` call together — or,
short of that, an extracted, separately unit-testable function at each
call site that a test can assert calls `canonical_workspace_path`.

Filed from #1040's multi-axis review, 2026-09-22.

## Blocked by

- #1040

PR #1088, opened 2026-09-22
flow/lane: canonicalize workspace paths before the worker-record compare

### LINKED ISSUE #1040: Worker-record cleanup match is symlink-fragile (exact path string)

caneff:
Closes #1040

## What changed

`canonical_workspace_path` (new, `flow/lane/src/workers.rs`) canonicalizes a
workspace path with `std::fs::canonicalize`, falling back to the string
verbatim when that fails. `implement-dispatch` runs its constructed
workspace path through it before storing it in a `WorkerRecord`, and
`merge-cleanup` runs the git-reported path through it before calling
`remove_workspace` — computed while the worktree directory still exists,
before `git worktree remove --force` deletes it. Both binaries now compare
one spelling of a workspace path instead of two independently-built ones,
so a symlinked repo root no longer makes a worker's cleanup record survive
`merge-cleanup` silently.

## Tests run

`cd flow/lane && cargo test --lib` — 43 passed, 0 failed.
`bash tests/all.sh` — 85 suites passed.

## Decisions made

Full multi-axis review round 1 (standards, spec, correctness) plus one
verification pass. Every finding:

- S1 — leftover (judgement): `canonical_workspace_path` repeats the
  canonicalize-or-fall-back shape `merge_cleanup.rs`'s `resolved()` /
  `resolved_under()` already carry. A real dedup reaches three call sites
  outside this diff; left as a documented tradeoff.
- S2 — leftover (judgement): the doc's stated fallback can't fully honor
  its promise for a torn-down directory. An in-round attempt to close this
  (sha `1bf57f5`) was reverted (sha `fe7ed03`) after the verification pass
  found it broke the adjacent-fix rule's 20-line budget (44 changed lines)
  for a non-high finding — see C2 below.
- P1 — leftover (judgement): records written by a pre-fix binary keep the
  raw, un-canonicalized spelling and stay unreachable against a
  canonicalized query on a symlinked root. Not a regression; no migration
  or either-spelling match exists.
- P2 — leftover (judgement): the shared normalizer is convention, not
  structure — nothing stops a third call site from comparing raw strings
  again.
- C1 — filed: caneff/agent-skills#1087 (blocked by #1040). Correctness
  `CONFIRMED`, so high under the severity mapping: reverting both call
  sites back to their pre-fix form leaves the whole crate's test suite
  green, so nothing pins that either binary actually calls the helper.
- C2 — leftover (judgement, `PLAUSIBLE`): same substance as S2 — the
  canonicalize fallback reproduces #1040's silent-survival failure when
  the workspace directory is already gone off-disk (a symlinked root, an
  earlier half-finished cleanup, a WSL restart). The in-round fix (sha
  `1bf57f5`) was sound (own new test went red under mutation, full suite
  green) but oversized for an adjacent fix on a non-high finding, so it
  was reverted (sha `fe7ed03`) rather than kept past the budget the round
  is scoped to.
- C3 — leftover (judgement, `PLAUSIBLE`): same root cause as P1.

## Last reviewed sha

`1bf57f5` — the verification pass ran against this sha (round-1 fix plus
the since-reverted adjacent fix). The revert at `fe7ed03` is past it and
was not independently re-reviewed; it exactly reverses `1bf57f5`, so it
undoes only what the verification pass already read.

- Gate pass (head fe7ed03), 1 medium: pre-fix worker records keep the raw path spelling and miss a canonical key under a symlinked root. **Disputed: unreachable** — no repo on this box sits under a symlinked root, so every existing record's raw spelling already equals its canonical one; it is also the worker's P1 leftover restated. No second pass.


caneff:
## Codex adversarial pass (gate, head fe7ed03)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: the new canonical key is not backward-compatible with worker records already stored using the prior raw path spelling.

Findings:
- [medium] Pre-fix worker records survive cleanup under a symlinked root (flow/lane/src/workers.rs:164-165)
  Records written before this change retain the raw `wt.display()` spelling. Cleanup now supplies a canonicalized workspace key, but removal still uses exact equality, so after an upgrade a raw record created through a symlinked repo path cannot match the canonical key. Because removal runs after `git worktree remove`, it cannot reliably canonicalize the vanished record path afterward. The record remains silently and can later yield a cleanup instruction for a deleted workspace.
  Recommendation: Add a compatibility path before worktree deletion: resolve and remove/migrate both existing raw and canonical record spellings while the worktree exists, or persist a stable git-reported identity and support both legacy raw values and the new key during comparison.

PR #1091, opened 2026-09-22
workers.rs: read takes the sidecar lock so restore never sees a torn rewrite

### LINKED ISSUE #1044: workers.rs: read takes the sidecar lock so restore never sees a torn rewrite
`remove_workspace` in `flow/lane/src/workers.rs` truncates and rewrites a controller's `<pid>.workers.jsonl` under an exclusive `File::lock()`, but `read` (what `controller-restore` calls on `SessionStart`) is an unlocked `read_to_string`. A `/clear` that fires while `merge-cleanup` is between the truncate and the rewrite reads an empty file and restores nothing, with no signal — the recovery path fails in exactly the window it exists for. Fix: `read` takes the same lock (shared is enough) so it only observes a complete generation, or the rewrite goes through a temp file and atomic rename. Regression test: pause cleanup after truncation, run restore, assert the retained record is read rather than an empty file. Evidence: PR #1043's second Codex pass, finding [medium] "Restore reads a torn cleanup rewrite without taking the file lock" (workers.rs:93-98 at 66e16cb).

Filed from: PR #1043's final Codex pass (controller disposition), 2026-09-22.

## Blocked by

- #964

caneff:
Closes #1044

## What changed

`workers.rs::read` now opens its sidecar `<pid>.workers.jsonl` and takes a shared `File::lock_shared()` before reading, so it blocks until `remove_workspace`'s exclusive-locked truncate-then-rewrite completes rather than observing the file mid-truncate. A lock failure is logged to stderr rather than silently falling back to an unlocked read. New regression test simulates the torn-truncate window directly and proves `read` blocks on it.

## Tests run

`cd flow/lane && cargo test --lib workers::` — 13 passed.
`bash tests/all.sh` (full repo gate) — all suites pass except `tests/lane-unparked.test.sh`, confirmed flaky and unrelated to this diff (fails and passes identically with the change stashed out and restored).

Mutation check: reverting `read` to the old unlocked `read_to_string`, and separately deleting only the `lock_shared()` call, both turn the new test red at its own assertion (`read ran without the lock`) — confirmed by the correctness axis reviewer in an isolated worktree.

## Decisions made

Every round-1 finding, by id (sidecars in `~/.cache/agent-reviews/skills/`):

- S1/P1/C2 (same finding, three axes — swallowed `lock_shared()` error, indistinguishable from the ordinary torn-rewrite failure `#1044` removes): `fixed (adjacent)`, d279147.
- S2 (needless `std::io::Read::` qualification): `fixed (adjacent)`, d279147.
- S3/C4 (same finding — the new test's first assertion is wall-clock-based): `disputed` — matches this file's own established precedent (`append_and_remove_serialize_on_the_file_lock_so_a_racing_append_is_never_lost` uses the identical `recv_timeout`/`is_err()` technique for the same class of claim); verified by the verification pass reading that precedent directly.
- P2 (the new test hand-rolls the truncate+rewrite rather than driving it through the real `remove_workspace`): `leftover` — the verification pass read the precedent test cited in the worker's dispute and found it does exercise the real `remove_workspace`/`append` on real threads (only the lock-holder is hand-rolled), which is not this test's shape; reclassified from the worker's claimed `disputed` to `leftover` since the finding is judgement-severity, not high.
- C1 (PLAUSIBLE, judgement — `read` now blocks unbounded on a wedged lock holder, bounded only by the SessionStart hook's own timeout): `leftover` — a retry/backoff fix is real new behavior beyond the adjacent-fix budget.
- C3 (judgement — mutating the sole real call site, `controller_restore.rs:220`, to bypass the lock left the whole crate suite green, so no test protects that call site): `leftover` — closing it needs integration coverage on the `controller-restore` binary, a second file, beyond the adjacent-fix budget.

## Last reviewed sha

d279147 — no commits past the reviewed sha.

- Gate pass (head d279147), medium 1: a failed `lock_shared()` falls back to the unlocked read. **Disputed: unreachable** — `flock` on a local ext4 file under WSL does not fail (ENOLCK is an NFS error), and the fail-closed rewrite needs a new return type plus a second file, outside the adjacent-fix budget; the worker checked and declined correctly.
- Gate pass, medium 2: the race test's 300 ms window may not prove the reader reached the lock. **Disputed**: the worker's mutation check (lock call deleted) turned the test red at its own assertion, and the technique is this file's own precedent.
- Head unmoved; no second pass.


caneff:
## Codex adversarial pass (gate, head d279147)

# Codex Adversarial Review

Verdict: needs-attention

No-ship: the lock-error fallback reintroduces the torn-read data-loss path, and its regression test can pass without exercising the lock.

Findings:
- [medium] A failed shared-lock acquisition deliberately reads the file during a rewrite (flow/lane/src/workers.rs:119-125)
  If `lock_shared()` fails after `remove_workspace` has acquired its exclusive lock and truncated the file, this path logs and immediately performs the previously-vulnerable unlocked read. It can therefore return an empty worker set to `controller-restore`, losing recovery output in the exact partial-write window the change is meant to close. Logging makes this diagnosable but does not preserve the recovery invariant.
  Recommendation: Do not consume the file unlocked after lock acquisition fails. Propagate a distinguishable read/lock error to `controller-restore` so it can report and retry/fail explicitly, or redesign the rewrite as an atomic publication that makes unlocked readers observe a complete generation.
- [medium] The race regression test does not prove the reader reached the lock (flow/lane/src/workers.rs:397-410)
  The 300 ms timeout only proves that `read` has not returned; it does not prove that the spawned thread ran far enough to invoke `lock_shared()`. If the thread is merely delayed until after `held` is dropped, the assertion and final full-record assertion both pass even if the lock call is removed. Thus the required regression can pass while the original torn-read bug is back.
  Recommendation: Add deterministic synchronization at the lock-acquisition boundary (for example a test-only hook that signals immediately before `lock_shared()`), then require the reader to reach that boundary before asserting it remains blocked until the rewrite completes.

Next steps:
- Make lock acquisition failure fail closed with an explicit recovery-path signal, then replace the scheduler-dependent test with a boundary-synchronized race test.


Controller: finding 1 confirmed, one adjacent fix round sent (fail closed on a lock error). Finding 2 disputed: the worker's mutation check turned the test red with the lock call removed, and the timing technique is this file's own precedent.

